PR #6, opened 2016-08-04
Implement URL alias view action override through semantic config

pspanja:


pspanja:
Ping @emodric

bundle/DependencyInjection/Configuration.php
```diff
@@ -0,0 +1,39 @@
...
+        $this->addConfiguration($rootNode);
+
+        return $treeBuilder;
+    }
+
+    protected function addConfiguration($rootNode)
+    {
+        $systemNode = $this->generateScopeBaseNode($rootNode);
+        $systemNode
+            ->booleanNode('override_url_alias_view_action')
```
emodric, line 31: How about just `override_default_controller` ?

emodric:
Other than the inline comment, +1.

bundle/DependencyInjection/Configuration.php
pspanja, line 31: The thing is there is no such concept in the code, this actually **is** the action when the view is resolved through URL alias.

pspanja:
Thanks!

PR #8, opened 2016-08-04
Update value objects to more closely mimic original value objects

emodric:
This allows less code modifications once the Site API is activated.

Once the override for default controller action is activated, this also saves time by requiring less changes to Twig templates.

Included in PR also:
- Renamed `$field->identifier` to `$field->fieldDefIdentifier` for above reason
- Added `$location->contentId` for above reason
- Added convenience properties to fetch translated content type name and content type description in content info

TODO:
- [x] Update tests


emodric:
Ping @pspanja

pspanja:
Thanks @emodric!

PR #9, opened 2016-08-29
Override preview controller if configured

emodric:


emodric:
Rebased on top of master after merging #11

pspanja:
Thanks @emodric!

PR #11, opened 2016-08-29
Remove possibility to configure view action

emodric:
Overriding the view action can effectively disable the content view builder due
to how matching is done on which view builder will be used (it checks for hardcoded `ng_content:` prefix).

After merging, https://github.com/netgen/ezplatform-site-api/pull/9 needs to be rebased.


pspanja:
Thanks @emodric!

PR #14, opened 2016-08-29
Add base controller which other controllers can extend

emodric:
When writing controllers, I find myself always injecting find service and load service over and over again.

This adds a base controller much like eZ and Symfony base controllers with helper method to load site API.

Also overriden is the method to load root location, which returns site API location now.


pspanja:
+1

PR #16, opened 2016-08-29
Node is based on Location instead of Content

pspanja:


pspanja:
Ping @emodric @mivancic

emodric:
+1

PR #22, opened 2016-12-12
Allow eZ value objects to be transfered to ng_content:viewAction

emodric:
They will be automatically converted to Site API value objects.

In case eZ Location value object is provided, Site API Location value object is reloaded without checking for visibility, to maintain how the original code works, which does not check for visibility if Location value object is already provided in the request.


pspanja:
Looks good, thanks @emodric!

emodric:
Thanks!

MarioBlazek:
Nice one :+1:

PR #23, opened 2017-01-17
Use custom view providers, to be able to use ngcontent_view override rules

emodric:
This adds the ability to use `ngcontent_view` template override rules when using Site API, by implementing custom view providers, to allow `ez_content:viewAction` to work without having to fallback to Site API specific templates:

You can now have the following config:

```
ezpublish:
    system:
        frontend_group:
            ngcontent_view:
                full:
                    frontpage:
                        template: "NetgenSiteBundle:content/full/siteapi:frontpage.html.twig"
                        match:
                            Identifier\ContentType: frontpage
            content_view:
                full:
                    frontpage:
                        template: "NetgenSiteBundle:content/full:frontpage.html.twig"
                        match:
                            Identifier\ContentType: frontpage
```

To facilitate this, Site API `ContentView` does not extend from base `ContentView` any more, since view providers are selected based on view object class and extending from base view would mean that original view providers would be used instead of our custom ones.

It is also backwards compatible (aside from changes to `ContentView` object and it's interfaces), meaning, if `ngcontent_view` rules do not exist, Site API rendering will fallback to `content_view` rules.

### DIFF
--- a/bundle/DependencyInjection/Configuration/Parser/ContentView.php
+++ b/bundle/DependencyInjection/Configuration/Parser/ContentView.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Netgen\Bundle\EzPlatformSiteApiBundle\DependencyInjection\Configuration\Parser;
+
+use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\Parser\ContentView as BaseContentView;
+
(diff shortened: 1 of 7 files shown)


MarioBlazek:
Nice one, +1

pspanja (approved):
Awesome :)

PR #26, opened 2017-02-09
PHPUnit forward compatibility layer

MarioBlazek:


MarioBlazek:
That would be big fall down the cliff, thanks @emodric , I did not thought about that. I'm closing PR.

PR #27, opened 2017-03-03
Mark this bundle as ezplatform-bundle

MarioBlazek:
Makes it possible to query ezplatform bundles via packagist: https://packagist.org/packages/list.json?type=ezplatform-bundle


pspanja:
Nice, thanks @MarioBlazek!

PR #28, opened 2017-03-03
Autoload tests only for dev

MarioBlazek:
This PR makes use of composer autoload-dev feature. It will autoload tests only in dev environment.


pspanja (approved):

pspanja:
Thanks @MarioBlazek

PR #29, opened 2017-07-15
Site API goes for lazy loading

pspanja:
This implements new lazy-loaded properties and methods on Site API objects, enabling simple content tree traversal without boilerplate code.

List of new properties by example in Twig:

1. Content and ContentInfo can access their Locations:
  `{% for location in content.locations %}...`
  `{% for location in contentInfo.locations %}...`

2. Location and Node can access their children Locations
  `{% for child in location.children %}...`
  `{% for child in node.children %}...`

3. Location and ContentInfo can access their Content
  `{{ location.content.name }}`
  `{{ contentInfo.content.fields.title.value.text }}`

4. Location and Node can access their parent Location
  `{{ location.parent.contentInfo.name }}`
  `{{ node.parent.contentInfo.name }}`

5. Location and Node can access their sibling Locations (does not include current Location)
  `{% for sibling in location.siblings %}...`
  `{% for sibling in node.siblings %}...`

6. Content and ContentInfo can access their main Location
  `{% for child in content.mainLocation.children %}...`
  `{% for child in contentInfo.mainLocation.children %}...`

List of new methods on Location and Node:

`{% for child in location.getChildren(5) %}...`
`{% for child in location.getChildren(5, ['article]) %}...`
`{% for sibling in location.getSiblings(5) %}...`
`{% for sibling in location.getSiblings(5, ['article]) %}...`

Note that since it's now possible to access Content from Location, Node object is deprecated.

### DIFF
--- a/lib/Core/Site/DomainObjectMapper.php
+++ b/lib/Core/Site/DomainObjectMapper.php
@@ -2,8 +2,6 @@
 
 namespace Netgen\EzPlatformSiteApi\Core\Site;
 
-use Netgen\EzPlatformSiteApi\API\FindService as APIFindService;
-use Netgen\EzPlatformSiteApi\API\LoadService as APILoadService;
 use Netgen\EzPlatformSiteApi\Core\Site\Values\Content;
 use Netgen\EzPlatformSiteApi\Core\Site\Values\ContentInfo;
 use Netgen\EzPlatformSiteApi\Core\Site\Values\Location;
--- a/lib/Core/Site/DomainObjectMapper.php
+++ b/lib/Core/Site/DomainObjectMapper.php
@@ -2,6 +2,8 @@
 
 namespace Netgen\EzPlatformSiteApi\Core\Site;
 
+use Netgen\EzPlatformSiteApi\API\FindService;
+use Netgen\EzPlatformSiteApi\API\LoadService;
 use Netgen\EzPlatformSiteApi\Core\Site\Values\Content;
 use Netgen\EzPlatformSiteApi\Core\Site\Values\ContentInfo;
 use Netgen\EzPlatformSiteApi\Core\Site\Values\Location;
@@ -33,6 +35,16 @@ final class DomainObjectMapper
     private $contentTypeService;
 
     /**
+     * @var \Netgen\EzPlatformSiteApi\API\FindService
+     */
+    private $findService;
+
+    /**
+     * @var \Netgen\EzPlatformSiteApi\API\LoadService
+     */
+    private $loadService;
+
+    /**
      * @param \eZ\Publish\API\Repository\ContentTypeService $contentTypeService
      * @param \eZ\Publish\API\Repository\FieldTypeService $fieldTypeService
      */
@@ -42,6 +54,22 @@ final class DomainObjectMapper
     ) {
         $this->contentTypeService = $contentTypeService;
         $this->fieldTypeService = $fieldTypeService;
+    }
+
+    /**
+     * @param \Netgen\EzPlatformSiteApi\API\FindService $findService
+     */
+    public function setFindService(FindService $findService)
+    {
+        $this->findService = $findService;
+    }
+
+    /**
+     * @param \Netgen\EzPlatformSiteApi\API\LoadService $loadService
+     */
+    public function setLoadService(LoadService $loadService)
+    {
+        $this->loadService = $loadService;
     }
 
     /**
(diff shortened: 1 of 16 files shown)


MarioBlazek:
This is great addition to site api, thanks @pspanja.

lib/Core/Site/Values/Content.php
```diff
@@ -143,4 +176,42 @@ private function buildField(array $properties = [])
         $this->fields[$field->fieldDefIdentifier] = $field;
         $this->fieldsById[$field->id] = $field;
     }
+
+    private function getLocations()
```
slaci, line 180: This method should have an optional `$limit` parameter I think (like `getChildren()`). In the current form it will limit to 25 locations, no? Well 25 is usually more than enought for 1 content, but sometimes not I think.

pspanja: Hi @slaci, you're right, I will make it public and add the parameter.
I also missed to mention it in the docblock and the condition to include only visible Locations, will have to fix that as well.

pspanja: Added in https://github.com/netgen/ezplatform-site-api/commit/654be5ffd01d4f1f4f74246020e2e9e899a4e7f4

Any preference on how these should be sorted? It should be done at least for consistence. I was thinking depth then path. Priority is another option, but I think it makes sense only when used among siblings.

slaci: For our last project we had to sort everything by Priority ASC, but the main point was that to make it consistent with the admin, but thats not the default sort for the admin either I think, so I think its too subjective.
I would vote for some simple sorting, dunno how your double level sort affects performance, but I guess its a bit overhead for eZ.
The best would be to make it customizable by a param, but then the `internalLocations` cache could cause problems withouth making that complex :)

pspanja: In administration these are sorted by path string, so I've done the same here, in https://github.com/netgen/ezplatform-site-api/commit/5ebead06c230721ac825611f5d8dc702ed9de07c.

I don't want to make this too elaborate, the aim is only to enable doing trivial traversal from the template. But if there's a clear need for that in the future, we might add it anyway.

slaci: That is more than enought, for custom stuff everyone can make query types.

lib/Core/Site/DomainObjectMapper.php
```diff
@@ -22,6 +23,8 @@
  */
 final class DomainObjectMapper
 {
+    use ContainerAwareTrait;
```
emodric, line 26: Is there a reason why you didn't inject the container through the constructor?

Or even inject the site with a setter in order not to use the container at all?

lib/Core/Site/Values/Content.php
```diff
@@ -36,6 +54,11 @@ public function __construct(array $properties = [])
             unset($properties['_fields_data']);
         }
 
+        if (array_key_exists('site', $properties)) {
+            $this->site = $properties['site'];
```
emodric, line 58: This is handled by the parent constructor, so the entire `if` here is superfluous, I think.

MarioBlazek: @emodric yep, base value object constructor will handle this.

pspanja: It has to be this way, this extends from eZ's [`ValueObject`](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/API/Repository/Values/ValueObject.php), and `$site` property here is private. So I can't pass it to the parent constructor because it would fail in the magic setter for not being visible there, and if I were to make it `protected`, it would be readable, which is a no-go.

lib/Core/Site/DomainObjectMapper.php
pspanja: I can't inject the Site directly because of circular dependency mess, it would still be called upon instantiation. If you have a suggestion how to get rid of that without exploding everything in the process, I'm all ears :)

I can inject the container through the constructor if you prefer that, it should also work.

emodric: Even if it's injected with a setter? Didn't try it myself, so that was just a random thought on top of my head.

As for using the trait, you can leave it, I was just wondering if there was a reason for using it.

emodric:
:+1:

lib/Core/Site/DomainObjectMapper.php
pspanja: Yeah, just tried it, it doesn't really make a difference.

PR #31, opened 2017-07-26
Introduce FilterService

pspanja:
This introduces `FilterService`, which is mostly the same as `FindService`, only it uses Legacy Search Engine, which means it's synchronous, that is data does not need to be indexed in order to be available in search.

Lazy loaded properties and methods are changed to use `FilterService` instead of `FindService`.

Note that fulltext search with `FilterService` will not work, because Legacy Search Engine slots will not be activated here (of course -- unless you also configure `legacy` search engine with Repository). The main idea behind this is to make synchronous search possible when using Solr Search Engine.

Also note that fetching `Node` objects is not possible through `FilterService`, since `Node` was previously deprecated.

### DIFF
--- a/lib/Resources/config/internal.yml
+++ b/lib/Resources/config/internal.yml
@@ -12,6 +12,16 @@ services:
         calls:
             - [setContainer, ['@service_container']]
 
+    netgen.ezplatform_site.core.filter_service:
+        class: Netgen\EzPlatformSiteApi\Core\Site\FilterService
+        lazy: true
+        arguments:
+            - '@netgen.ezplatform_site.core.domain_object_mapper'
+            - '@netgen.ezplatform_site.repository.filter_service'
+            - '@ezpublish.api.service.content'
+            - '%netgen.ezplatform_site.prioritized_languages%'
+            - '%netgen.ezplatform_site.use_always_available%'
+
     netgen.ezplatform_site.core.find_service:
         class: Netgen\EzPlatformSiteApi\Core\Site\FindService
         lazy: true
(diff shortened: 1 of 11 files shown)


lib/Resources/config/internal.yml
```diff
@@ -36,5 +46,24 @@ services:
...
+        factory:
+            - '@ezpublish.api.repository.factory'
+            - buildRepository
+        arguments:
+            - '@ezpublish.api.persistence_handler'
+            - '@ezpublish.spi.search.legacy'
+        lazy: true
+        public: false
+
+    netgen.ezplatform_site.repository.filter_service:
```
emodric, line 64: This can also be `public: false`

emodric (approved):

emodric:
Build fails on PHP 5 with `PHP Fatal error:  Cannot use Netgen\EzPlatformSiteApi\API\FilterService as FilterService because the name is already in use in /home/travis/build/netgen/ezplatform-site-api/lib/Core/Site/Site.php on line 5`, but aside from that, :+1:

pspanja:
@emodric my bad, should be fixed in https://github.com/netgen/ezplatform-site-api/pull/31/commits/962af1c2a266c4fa6f1b0d67a8913e163637507e

MarioBlazek (approved):

lib/Resources/config/internal.yml
pspanja: Added in https://github.com/netgen/ezplatform-site-api/pull/31/commits/6f9f085b2a203a2ef15fd2a5f294d1134b794eb8

emodric:
Go go go! :racing_car:

pspanja:
All mrđed now.

PR #32, opened 2017-07-27
Introduce pagination

pspanja:
This introduces possibility of pagination, for now only on Location children.

`getChildren($limit = 25, array $contentTypeIdentifiers = [])` method was renamed and signature was changed to `filterChildren(array $contentTypeIdentifiers = [], $maxPerPage = 25, $currentPage = 1)`. The method now returns `Pagerfanta` instance.

Note that we don't have offset as initially discussed, as it would have to be converted to current page anyway. Comments welcome on that part.

When `children` property is accessed, internal iterator is returned instead of `Pagerfanta` instance.

This enables following:

```twig
{% set children = location.parent.filterChildren %}
<p>Total count: {{ children.nbResults }}</p>
<p>Page: {{ children.currentPage }}</p>
<p>Max per page: {{ children.maxPerPage }}</p>
<ul>
{% for child in children %}
    <li>{{ child.contentInfo.name }}</li>
{% endfor %}
</ul>
{{ pagerfanta( children, 'twitter_bootstrap' ) }}
```

Note the introduction of `LocationSearchFilterAdapter` adapter for `Pagerfanta`, using `FilterService`.

### To-dos
- [x] expand this on the rest of the methods (locations, siblings)
- [x] ~~introduce `PagerBuilderTrait` to ease using pagination~~ left for some future PR

### DIFF
--- a/lib/API/Values/Content.php
+++ b/lib/API/Values/Content.php
@@ -17,7 +17,6 @@ use eZ\Publish\API\Repository\Values\ValueObject;
  * @property-read \Netgen\EzPlatformSiteApi\API\Values\ContentInfo $contentInfo
  * @property-read \eZ\Publish\API\Repository\Values\Content\VersionInfo $versionInfo
  * @property-read \Netgen\EzPlatformSiteApi\API\Values\Field[] $fields
- * @property-read \Netgen\EzPlatformSiteApi\API\Values\Location[] $locations Up to 25 Content Locations
  * @property-read \Netgen\EzPlatformSiteApi\API\Values\Location|null $mainLocation
  * @property-read \eZ\Publish\API\Repository\Values\Content\Content $innerContent
  */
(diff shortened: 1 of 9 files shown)


pspanja:
Description updated.

MarioBlazek (approved):

emodric (approved):

hknezevic (approved):

pspanja:
In https://github.com/netgen/ezplatform-site-api/pull/32/commits/416b427a3fd0f1e079c897344f9609566d00c44c I introduced `getLocations()`, `getChildren()` and `getSiblings()` and removed `$locations`, `$children` and `$siblings` lazy loaded parameters.

The reason for that is to make trivial (`get*`) and more complex (`filter*`) usage more visually distinct.
Also, in Twig, `get*` methods can be accessed as properties without `get` part, so it will still be possible to do for example `{% for child in location.children %} ...`.

Compared to that `{% for child in location.filterChildren(...) %} ...` hopefully will stand apart and thus reduce cognitive load when reading code :)

PR #33, opened 2017-07-27
Display scrutinizer score

MarioBlazek:


pspanja:
Thanks @MarioBlazek 👍

PR #35, opened 2017-07-27
Refactor service building

pspanja:
This refactors how services are built, replacing explicit configuration with using `Site` as factory.

Also, prioritized languages are now injected to the `Site` through setter method, which should enable updating on scope change.

### DIFF
--- a/lib/Core/Site/Site.php
+++ b/lib/Core/Site/Site.php
@@ -104,58 +104,55 @@ class Site implements SiteInterface
      *
      * @param array $prioritizedLanguages
      */
-    public function setPrioritizedLanguages(array $prioritizedLanguages)
+    public function setPrioritizedLanguages(array $prioritizedLanguages = null)
     {
         $this->prioritizedLanguages = $prioritizedLanguages;
     }
 
     public function getFilterService()
     {
-        if ($this->filterService !== null) {
-            return $this->filterService;
+        if ($this->filterService === null) {
+            $this->filterService = new FilterService(
+                $this->getDomainObjectMapper(),
+                $this->synchronousSearchService,
+                $this->contentService,
+                $this->useAlwaysAvailable
+            );
         }
 
-        $this->filterService = new FilterService(
-            $this->getDomainObjectMapper(),
-            $this->synchronousSearchService,
-            $this->contentService,
-            $this->prioritizedLanguages,
-            $this->useAlwaysAvailable
-        );
+        $this->filterService->setPrioritizedLanguages($this->prioritizedLanguages);
 
         return $this->filterService;
     }
 
     public function getFindService()
     {
-        if ($this->findService !== null) {
-            return $this->findService;
+        if ($this->findService === null) {
+            $this->findService = new FindService(
+                $this->getDomainObjectMapper(),
+                $this->searchService,
+                $this->contentService,
+                $this->useAlwaysAvailable
+            );
         }
 
-        $this->findService = new FindService(
-            $this->getDomainObjectMapper(),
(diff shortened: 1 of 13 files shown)


lib/Core/Site/Site.php
```diff
@@ -25,32 +72,110 @@ class Site implements SiteInterface
...
+        $this->useAlwaysAvailable = $useAlwaysAvailable;
+    }
+
+    /**
+     * Setter for prioritized languages configuration, used to update
+     * the configuration on scope change.
+     *
+     * @param array $prioritizedLanguages
+     */
+    public function setPrioritizedLanguages(array $prioritizedLanguages)
```
emodric: Should be `null` by default, as per docs.

lib/Core/Site/Site.php
```diff
@@ -25,32 +72,110 @@ class Site implements SiteInterface
...
     {
+        if ($this->filterService !== null) {
+            return $this->filterService;
+        }
+
+        $this->filterService = new FilterService(
+            $this->getDomainObjectMapper(),
+            $this->synchronousSearchService,
+            $this->contentService,
+            $this->prioritizedLanguages,
```
emodric: This should also be injected (on other services too), otherwise changing scope will not work.

pspanja: Right, but then it seems this approach with `Site` as factory is a no go.
Or maybe we can add `calls` on the factory anyway.

emodric: Yep, l know. I'm not really sure what to do here. We didn't have any issues so far, so maybe it's not worth the trouble to use setter injection.

pspanja: Well, here's the solution, though I'm not sure what to do with it, maybe it should be preventively burned, just in case: https://github.com/netgen/ezplatform-site-api/pull/35/commits/bd36bb85efd2e473fe935152c6f7c1397bde5074

emodric: With that, `setPrioritizedLanguages` on `Site` instance should not be needed, right? It will be called on other three services when instantianting them, right?

pspanja: It's still needed if you don't fetch the services from the container, but from the `Site`. In that case setter would not be called by the container.

emodric: Ah, you're right. It makes me wondering, do we really need `Site` instance at all?

pspanja: So far it only aggregated other services, so not really, though it's convenient to inject only one service instead of three.

But now with this PR it's also used as a factory. That is it builds services in a way that does not require injecting container to the domain mapper, because it has circular dependency on other API services. That's what I tried to get rid of here.

pspanja: I think I'm going to implement `LanguageSettingsProvider` I mentioned privately, if only to see how it looks :)

But thinking more about it, maybe this is something to be generalized and recognized as a first class citizen in Site API? `SettingsProvider` or something.

emodric: It would wrap config resolver then to provide only settings available in site API?

pspanja: Yes, I think that would be the most sensible approach.

pspanja:
Added API `Settings` to centralize all used settings in https://github.com/netgen/ezplatform-site-api/pull/35/commits/fff2779dbacf09716cb57049f4e4f5212d1254ae.
It uses `ConfigResolver`, so settings should now be in sync when scope changes.

emodric:
:+1:

MarioBlazek (approved):

PR #40, opened 2017-08-01
Unit and integration composer commands

MarioBlazek:
Added two composer command, for running unit and integration tests.

Unit tests:
```bash
composer test
```

Integration tests:
```bash
composer test-integration
```


pspanja:
Thanks @MarioBlazek 👍

PR #41, opened 2017-08-01
PHPUnit forward compatibility layer

MarioBlazek:


MarioBlazek:
@pspanja WDYT ?

composer.json
```diff
@@ -14,7 +14,7 @@
         "ezsystems/ezpublish-kernel": "^6.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "~4.7"
+        "phpunit/phpunit": "^4.8.35|^5.7"
```
pspanja, line 17: Do we need `4.8.35` for some reason, shouldn't `^4.8|^5.7` be enough?

MarioBlazek: in 4.8.35 `TestCase` class was introduced.

MarioBlazek: https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4835---2017-02-06

pspanja: OK then :)

pspanja:
Looks good, thanks @MarioBlazek!

PR #42, opened 2017-08-01
[2.3] Introduce simple forward relations

pspanja:
This introduces `RelationService` with a (for now) single method:

`loadFieldRelations($contentId, $fieldIdentifier, array $contentTypeIdentifiers = [])`

The method will return **all** forward related Content of a given field.
Note:
1. No filtering on Location visibility is done
2. No offset or limit is provided as with current implementation we need to load all relations anyway

In future that might change as we might implement search (filtering) support for forward and reverse relations and something called `LocationQueryCriterion` to enable grouping of Location conditions in Content search.

On the `Content` object two new methods are added:
- `getFieldRelations($fieldDefinitionIdentifier)`
- `filterFieldRelations($fieldDefinitionIdentifier, array $contentTypeIdentifiers = [])`

For Twig the usage looks like this:

```twig
<ul>
{% for wine in content.fieldRelations('wine') %}
    <li>{{ wine.name }}</li>
{% endfor %}
</ul>
```

```twig
<ul>
{% for cheese in content.filterFieldRelations('cheese', ['cheese'])|slice(0, 5) %}
    <li>{{ cheese.name }}</li>
{% endfor %}
</ul>
```

### DIFF
--- a/lib/API/RelationService.php
+++ b/lib/API/RelationService.php
@@ -8,8 +8,30 @@ namespace Netgen\EzPlatformSiteApi\API;
 interface RelationService
 {
     /**
-     * Load related Content from $fieldDefinitionIdentifier field in Content with given $contentId,
-     * optionally limited by a list of $contentTypeIdentifiers.
+     * Load single related Content from $fieldDefinitionIdentifier field in Content with given
+     * $contentId, optionally limited by a list of $contentTypeIdentifiers.
+     *
+     * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentValue If the $fieldDefinitionIdentifier does
+     *         not exist in the Content with given $contentId
+     *
+     * @param $contentId
+     * @param $fieldDefinitionIdentifier
+     * @param array $contentTypeIdentifiers
+     *
+     * @return \Netgen\EzPlatformSiteApi\API\Values\Content|null
+     */
+    public function loadFieldRelation(
+        $contentId,
+        $fieldDefinitionIdentifier,
+        array $contentTypeIdentifiers = []
+    );
+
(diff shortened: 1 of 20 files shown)


MarioBlazek:
@pspanja why shouldn't  we provide support `ezobjectrelation` field type along side `ezobjectrelationlist` ?

MarioBlazek:
Also we should add Location visibility check and at least option to limit number of returned Content objects. Besides that you have +1

lib/API/RelationService.php
```diff
@@ -0,0 +1,21 @@
...
+     * Load related Content from $fieldDefinitionIdentifier field in Content with given $contentId,
+     * optionally limited by a list of $contentTypeIdentifiers.
+     *
+     * @param int|string $contentId
+     * @param string $fieldDefinitionIdentifier
+     * @param array $contentTypeIdentifiers
+     *
+     * @return \Netgen\EzPlatformSiteApi\API\Values\Content[]
+     */
+    public function loadFieldRelations($contentId, $fieldDefinitionIdentifier, array $contentTypeIdentifiers = []);
```
MarioBlazek: Maybe to add additional method `loadFieldRelation` that will handle single relation ?

MarioBlazek: I suggest to have `limit` parameter.

MarioBlazek (approved):

pspanja:
@MarioBlazek now updated per your spec from https://github.com/MarioBlazek/specs/blob/master/Relation_API_spec.md
This should be usable, I'll try to test it on a real use-case in next few days.

emodric:
:tada:

PR #50, opened 2017-10-05
Introduce __debugInfo for cleaner dump

pspanja:
This introduces [__debugInfo()](http://php.net/manual/en/language.oop5.magic.php#object.debuginfo) on the domain objects, in order to have saner output when dumping them for debugging.

This is to avoid dumping aggregated services, but also some of the properties are omitted to avoid recursion and tree traversal (`$location->parent->parent...`, `$content->mainLocation->content...` etc). Some other properties (most of `inner*` bunch) are also omitted as it seems to me they usually won't be of interest, so would end up adding noise.

Skipped properties are commented out for the purpose of review.

Note: `__debugInfo()` is available only with PHP >=5.6.

### DIFF
--- a/lib/Core/Site/Values/Content.php
+++ b/lib/Core/Site/Values/Content.php
@@ -283,4 +283,21 @@ final class Content extends APIContent
 
         return $this->contentInfo;
     }
+
+    public function __debugInfo()
+    {
+        $this->initializeFields();
+
+        return [
+            'id' => $this->id,
+            'mainLocationId' => $this->mainLocationId,
+            'name' => $this->name,
+            'languageCode' => $this->languageCode,
+            'contentInfo' => $this->getContentInfo(),
+            'fields' => $this->fields,
+            //'mainLocation' => $this->getMainLocation(),
+            //'innerContent' => $this->getInnerContent(),
+            //'innerVersionInfo' => $this->innerVersionInfo,
+        ];
+    }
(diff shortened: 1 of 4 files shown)


MarioBlazek:
Woohoo, I like it :+1:

emodric:
👍 for removal of properties leading to recursion, but not sure if we should remove `inner*` properties. How much noise do they add really, considering we can use `dump()` PHP and Twig functions which handle the nice output for us?

Case in point, having `innerField` in debug output definitely helped while debugging #49 .

pspanja:
True, maybe we can keep them. @MarioBlazek what do you think?

pspanja:
@emodric I started adding those and there will be quite some duplication, for example:

- Repository fields in the `Content::$innerContent` and also in `Field::$innerField`
- FieldDefinition from `Content::$innerContentType` and from `Field::$innerFieldDefinition`
- ContentInfo from `Location::$innerLocation` and `Location::$contentInfo`

And so on. So I think I'm for leaving this as it is, particularly since it's easy to separately dump whatever is needed.
But I'll let @MarioBlazek decide, since he is the one using it :)

emodric:
Okay with me :)

emodric (approved):

pspanja:
Thanks @emodric and @MarioBlazek!

PR #51, opened 2017-10-05
Introduce $owner and $innerOwnerUser Content properties

pspanja:
This adds new lazy-loaded properties to `Content` object:

- `$owner`, instance of `\Netgen\EzPlatformSiteApi\API\Values\Content`
- `$innerOwnerUser`, instance of `\eZ\Publish\API\Repository\Values\User\User`

Since we can't be sure that Content's owner actually exists, `NotFoundException` is intercepted to return `null`.

### TODOs
- [x] add tests
- [x] handle `__debugInfo()` from #50 
- [x] properly hint that value can be `null`

### DIFF
--- a/lib/Core/Site/Values/Content.php
+++ b/lib/Core/Site/Values/Content.php
@@ -332,7 +332,7 @@ final class Content extends APIContent
         }
 
         try {
-            $this->owner = $this->site->getLoadService()->loadContent($this->contentInfo->ownerId);
+            $this->owner = $this->site->getLoadService()->loadContent($this->getContentInfo()->ownerId);
         } catch (NotFoundException $e) {
             // Do nothing
         }
@@ -352,7 +352,7 @@ final class Content extends APIContent
         }
 
         try {
-            $this->innerOwnerUser = $this->userService->loadUser($this->contentInfo->ownerId);
+            $this->innerOwnerUser = $this->userService->loadUser($this->getContentInfo()->ownerId);
         } catch (NotFoundException $e) {
             // Do nothing
         }
--- a/lib/Core/Site/Values/Content.php
+++ b/lib/Core/Site/Values/Content.php
@@ -98,6 +98,20 @@ final class Content extends APIContent
      */
     private $internalMainLocation;
 
+    /**
+     * Denotes if $owner property is initialized.
+     *
+     * @var bool
+     */
+    private $isOwnerInitialized = false;
+
+    /**
+     * Denotes if $innerOwnerUser property is initialized.
+     *
+     * @var bool
+     */
+    private $isInnerOwnerUserInitialized = false;
+
     public function __construct(array $properties = [])
     {
         $this->site = $properties['site'];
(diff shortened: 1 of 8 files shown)


pspanja:
Also ping @lolautruche, I think you suggested this on Web Summer Camp :) Any input is welcome.

emodric:
Looks good! 👍 

Was gonna suggest to switch `$innerOwnerUser` to just `$ownerUser`, but for consistency sake where all `$inner*` properties are repo value objects, it's better the way it is.

pspanja:
Yup, maybe should have gone with `$repo*` instead, but too late now :)

emodric (approved):

lib/Core/Site/Values/Content.php
```diff
@@ -283,4 +307,44 @@ private function getContentInfo()
...
+    /**
+     * @return \Netgen\EzPlatformSiteApi\API\Values\Content
+     */
+    private function getOwner()
+    {
+        if ($this->owner === null) {
+            try {
+                $this->owner = $this->site->getLoadService()->loadContent($this->contentInfo->ownerId);
+            } catch (NotFoundException $e) {
+                $this->owner = false;
```
lolautruche: Why assigning false? Null is much more meaningful there (i.e. "Nothing there"). A boolean value **is** an actual value, while null is a *non-value* :wink:

lib/Core/Site/Values/Content.php
```diff
@@ -283,4 +307,44 @@ private function getContentInfo()
...
+    /**
+     * @return \eZ\Publish\API\Repository\Values\User\User
+     */
+    private function getInnerOwnerUser()
+    {
+        if ($this->innerOwnerUser === null) {
+            try {
+                $this->innerOwnerUser = $this->userService->loadUser($this->contentInfo->ownerId);
+            } catch (NotFoundException $e) {
+                $this->innerOwnerUser = false;
```
lolautruche: Same remark as above. 
And you could get rid of the condition block below

lolautruche (changes requested):

lib/Core/Site/Values/Content.php
emodric: In order not to repeat the query if we already concluded that there's no owner user, I think.

lolautruche: Then better use return asap pattern (if user exists return it), and add a dedicated flag

pspanja: Yup, it's because it can happen that owner no longer exists, so it makes no sense to repeat loading if it was already done.

Updated as suggested in https://github.com/netgen/ezplatform-site-api/pull/51/commits/03952bc210937e9abc4649facec317120581b516, thanks.

lib/Core/Site/Values/Content.php
pspanja: Updated in https://github.com/netgen/ezplatform-site-api/pull/51/commits/03952bc210937e9abc4649facec317120581b516.

lib/Core/Site/Values/Content.php
lolautruche: 👌

pspanja:
I decided not to add new owner properties to the `__debugInfo()`, they probably won't be used very often so would mostly just add noise to the dump.

This should now be complete.

Edit: and conflicts are resolved as well :)

pspanja:
This looks perfectly ripe by now, thanks everyone 😊

PR #52, opened 2017-10-19
Added Pagerfanta Trait

MarioBlazek:
This PR adds shorthand method for creating Pagerfanta pagers.

Should reconsider:
* [x] method and trait naming
* [x] tests
* [x] updating methods documentation

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,6 +1,17 @@
 eZ Platform Site API upgrade instructions
 =========================================
 
+Upgrading from 2.2 to 2.3
+-------------------------
+
+Controllers that extend from `Netgen\Bundle\EzPlatformSiteApiBundle\Controller\Controller` and are registered inside dependency injection container must set `parent` config key within service definition, example:
+
+```yaml
+acme.demo.controller.demo_controller:
+    parent: netgen.ezplatform_site.controller.base
+    class: Acme\Bundle\DemoBundle\Controller\DemoController
+```
+
 Upgrading from 1.0 to 2.0
 -------------------------
 
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -4,10 +4,18 @@ eZ Platform Site API upgrade instructions
 Upgrading from 2.3 to 2.4
 -------------------------
 
-Controllers that extend from `Netgen\Bundle\EzPlatformSiteApiBundle\Controller\Controller` and are registered inside dependency injection container must set `parent` config key within service definition, example:
-
+Controllers that extend from `Netgen\Bundle\EzPlatformSiteApiBundle\Controller\Controller` and are registered inside dependency injection container should set two setter injection `calls`: 
 ```yaml
-acme.demo.controller.demo_controller:
+app.demo.controller.demo_controller:
+    class: Acme\Bundle\DemoBundle\Controller\DemoController
+    calls:
+        - [setContainer, ['@service_container']]
+        - [setSite, ['@netgen.ezplatform_site.site']]
+```
+
+Or if you want to avoid setter calls, just set `parent` service:
+```yaml
+app.demo.controller.demo_controller:
     parent: netgen.ezplatform_site.controller.base
     class: Acme\Bundle\DemoBundle\Controller\DemoController
 ```
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,7 +1,7 @@
 eZ Platform Site API upgrade instructions
 =========================================
 
-Upgrading from 2.2 to 2.3
+Upgrading from 2.3 to 2.4
 -------------------------
 
 Controllers that extend from `Netgen\Bundle\EzPlatformSiteApiBundle\Controller\Controller` and are registered inside dependency injection container must set `parent` config key within service definition, example:
--- a/bundle/Controller/Controller.php
+++ b/bundle/Controller/Controller.php
@@ -3,13 +3,13 @@
 namespace Netgen\Bundle\EzPlatformSiteApiBundle\Controller;
 
 use eZ\Bundle\EzPublishCoreBundle\Controller as BaseController;
-use Netgen\EzPlatformSiteApi\Core\Traits\PagerfantaFindServiceSearchAdaptersTrait;
+use Netgen\EzPlatformSiteApi\Core\Traits\PagerfantaFindTrait;
 use Netgen\EzPlatformSiteApi\Core\Traits\SearchResultExtractorTrait;
 
 abstract class Controller extends BaseController
 {
     use SearchResultExtractorTrait;
-    use PagerfantaFindServiceSearchAdaptersTrait;
+    use PagerfantaFindTrait;
 
     /**
      * @return \Netgen\EzPlatformSiteApi\API\Site
--- a/bundle/Controller/Controller.php
+++ b/bundle/Controller/Controller.php
@@ -12,14 +12,6 @@ abstract class Controller extends BaseController
     use PagerfantaFindTrait;
 
     /**
-     * @return \Netgen\EzPlatformSiteApi\API\Site
-     */
-    public function getSite()
-    {
-        return $this->container->get('netgen.ezplatform_site.site');
-    }
-
-    /**
      * Returns the root location object for current siteaccess configuration.
      *
      * @return \Netgen\EzPlatformSiteApi\API\Values\Location
--- a/bundle/Controller/Controller.php
+++ b/bundle/Controller/Controller.php
@@ -3,11 +3,13 @@
 namespace Netgen\Bundle\EzPlatformSiteApiBundle\Controller;
 
 use eZ\Bundle\EzPublishCoreBundle\Controller as BaseController;
+use Netgen\EzPlatformSiteApi\Core\Traits\PagerfantaFindServiceSearchAdaptersTrait;
 use Netgen\EzPlatformSiteApi\Core\Traits\SearchResultExtractorTrait;
 
 abstract class Controller extends BaseController
 {
     use SearchResultExtractorTrait;
+    use PagerfantaFindServiceSearchAdaptersTrait;
 
     /**
      * @return \Netgen\EzPlatformSiteApi\API\Site
--- a/lib/Core/Traits/PagerfantaFindServiceSearchAdaptersTrait.php
+++ b/lib/Core/Traits/PagerfantaFindServiceSearchAdaptersTrait.php
@@ -16,78 +16,74 @@ trait PagerfantaFindServiceSearchAdaptersTrait
 {
     /**
      * Returns Pagerfanta pager that starts from first page
-     * and uses 'page' GET parameter from Request object
      * configured with
      * ContentSearchAdapter and FindService
      *
      * @param Query $query
-     * @param Request $request
+     * @param int $currentPage
      * @param int $maxPerPage
      *
      * @return Pagerfanta
      */
-    protected function createContentSearchPager(Query $query, Request $request, $maxPerPage)
+    protected function createContentSearchPager(Query $query, $currentPage, $maxPerPage)
     {
         $adapter = new ContentSearchAdapter($query, $this->getSite()->getFindService());
 
-        return $this->getPagerfantaPager($adapter, $request, $maxPerPage);
+        return $this->getPagerfantaPager($adapter, $currentPage, $maxPerPage);
     }
 
     /**
      * Returns Pagerfanta pager that starts from first page
-     * and uses 'page' GET parameter from Request object
      * configured with
      * ContentSearchHitAdapter and FindService
      *
      * @param Query $query
-     * @param Request $request
+     * @param int $currentPage
      * @param int $maxPerPage
      *
      * @return Pagerfanta
      */
-    protected function createContentSearchHitPager(Query $query, Request $request, $maxPerPage)
+    protected function createContentSearchHitPager(Query $query, $currentPage, $maxPerPage)
     {
         $adapter = new ContentSearchHitAdapter($query, $this->getSite()->getFindService());
 
-        return $this->getPagerfantaPager($adapter, $request, $maxPerPage);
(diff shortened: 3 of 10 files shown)


lib/Core/Traits/PagerfantaFindServiceSearchAdaptersTrait.php
```diff
@@ -0,0 +1,113 @@
...
+     *
+     * @param AdapterInterface $adapter
+     * @param Request $request
+     * @param int $maxPerPage
+     *
+     * @return Pagerfanta
+     */
+    protected function getPagerfantaPager(AdapterInterface $adapter, Request $request, $maxPerPage)
+    {
+        $pager = new Pagerfanta($adapter);
```
emodric: This should also have `$pager->setNormalizeOutOfRangePages(true);`

lib/Core/Traits/PagerfantaFindServiceSearchAdaptersTrait.php
```diff
@@ -0,0 +1,113 @@
...
+        $pager->setMaxPerPage($maxPerPage);
+        $pager->setCurrentPage(
+            $request->query->get('page', 1)
```
emodric: Why not transfer the current page with the method parameter to remove dependency on the `Request` object?

lib/Core/Traits/PagerfantaFindServiceSearchAdaptersTrait.php
```diff
@@ -0,0 +1,113 @@
...
+use eZ\Publish\API\Repository\Values\Content\LocationQuery;
+use Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta\ContentSearchAdapter;
+use Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta\ContentSearchHitAdapter;
+use Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta\LocationSearchAdapter;
+use Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta\LocationSearchHitAdapter;
+use Pagerfanta\Adapter\AdapterInterface;
+use Pagerfanta\Pagerfanta;
+use Symfony\Component\HttpFoundation\Request;
+
+trait PagerfantaFindServiceSearchAdaptersTrait
```
emodric: `PagerfantaTrait` ? :)

MarioBlazek: In future maybe we would like to have trait that is internally using FilterService

lib/Core/Traits/PagerfantaFindServiceSearchAdaptersTrait.php
MarioBlazek: But why ? On all projects we are getting `page` parameter from `Request` object, with this it would be automatically done for you.

lib/Core/Traits/PagerfantaFindServiceSearchAdaptersTrait.php
MarioBlazek: Done.

lib/Core/Traits/PagerfantaFindServiceSearchAdaptersTrait.php
emodric: Then just `PagerfantaFindTrait`?

lib/Core/Traits/PagerfantaFindServiceSearchAdaptersTrait.php
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @return Pagerfanta
+     */
+    protected function getPagerfantaPager(AdapterInterface $adapter, Request $request, $maxPerPage)
+    {
+        $pager = new Pagerfanta($adapter);
+        $pager->setNormalizeOutOfRangePages(true);
+        $pager->setMaxPerPage($maxPerPage);
+        $pager->setCurrentPage(
+            $request->query->get('page', 1)
```
emodric: Is it a good idea to have the `Request` here? With this, I can't practically use the trait outside of controller code. It would be better to transfer the current page in method argument.

lib/Core/Traits/PagerfantaFindServiceSearchAdaptersTrait.php
emodric: Because this way, I can't just simply use the trait to create paged output outside of controller.

What if the query parameter is not named `page`?

What if I want to have two paged queries on the same page, one displaying a query in first page, one displaying a different query on second page?

It's not more difficult at all transfering `$request->query->get('page')` to the method call instead of just `$request`, code wise.

lib/Core/Traits/PagerfantaFindServiceSearchAdaptersTrait.php
MarioBlazek: Like it, +1, @pspanja wdyt ?

lib/Core/Traits/PagerfantaFindServiceSearchAdaptersTrait.php
MarioBlazek: I that case you should avoid using this shorthand methods. The idea is to cover 80-90% of use cases with this.

emodric: I disagree. The point of the trait is to be reusable in most usecases and using it outside of a controller where no request is readily available is a valid and often used usecase.

Moreover, using a page query parameter is only an implementation detail coming from Pagerfanta Bundle and not the Pagerfanta lib itself.

MarioBlazek: I've updated code. wdyt @emodric ?

emodric: Looks good :+1:

lib/Core/Traits/PagerfantaFindServiceSearchAdaptersTrait.php
pspanja: 👍

lib/Core/Traits/PagerfantaFindTrait.php
```diff
@@ -0,0 +1,107 @@
...
+use Pagerfanta\Pagerfanta;
+
+trait PagerfantaFindTrait
+{
+    /**
+     * Returns Pagerfanta pager that starts from first page
+     * configured with
+     * ContentSearchAdapter and FindService
+     *
+     * @param Query $query
```
pspanja: Please use FQCN in docblock hints.

pspanja:
I suggest reducing `create*Pager()` to only two methods, `createSearchPager()` and `createSearchHitPager()`. Inside them we can test given `$query` to decide what adapter to use.

lib/Core/Traits/PagerfantaFindTrait.php
```diff
@@ -0,0 +1,107 @@
...
+     * Shorthand method for creating Pagerfanta pager
+     * with preconfigured Adapter
+     *
+     * @param AdapterInterface $adapter
+     * @param int $currentPage
+     * @param int $maxPerPage
+     *
+     * @return Pagerfanta
+     */
+    protected function getPagerfantaPager(AdapterInterface $adapter, $currentPage, $maxPerPage)
```
pspanja: Given we don't have `Pagerfanta` in the method names above, maybe we can drop it here as well.
Also, this method can start with `create` like the ones above, or we can use `new` instead, like is done in Repository API. What do you think?

emodric: `getPager` sounds okay. I would keep it at `get*` or `new*` to keep it distinct from `create*` methods.

MarioBlazek:
Please review @pspanja @emodric

tests/lib/Integration/Traits/PagerfantaFindTraitTest.php
```diff
@@ -0,0 +1,65 @@
...
+use Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta\ContentSearchAdapter;
+use Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta\ContentSearchHitAdapter;
+use Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta\LocationSearchHitAdapter;
+use Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta\LocationSearchAdapter;
+use Netgen\EzPlatformSiteApi\Core\Traits\PagerfantaFindTrait;
+use Netgen\EzPlatformSiteApi\Tests\Integration\BaseTest;
+
+class PagerfantaFindTraitTest extends BaseTest
+{
+    use PagerfantaFindTrait;
```
emodric: This feels wrong. A test class that uses the trait under test instead of instantiating a (stub) object that uses the trait?

MarioBlazek:
@emodric wdyt ?

MarioBlazek:
Trait methods are using `$this->getSite()->getFindService()` calls.

emodric:
Shouldn't then `getSite` method as well as the `$site` property be defined inside the trait?

emodric:
Or ideally, only the property, without the method at all.

UPGRADE.md
```diff
@@ -1,6 +1,17 @@
 eZ Platform Site API upgrade instructions
 =========================================
 
+Upgrading from 2.2 to 2.3
+-------------------------
+
+Controllers that extend from `Netgen\Bundle\EzPlatformSiteApiBundle\Controller\Controller` and are registered inside dependency injection container must set `parent` config key within service definition, example:
```
pspanja: Why is this needed?

bundle/Controller/Controller.php
```diff
@@ -3,19 +3,13 @@
...
 use Netgen\EzPlatformSiteApi\Core\Traits\SearchResultExtractorTrait;
 
 abstract class Controller extends BaseController
 {
     use SearchResultExtractorTrait;
-
-    /**
-     * @return \Netgen\EzPlatformSiteApi\API\Site
-     */
-    public function getSite()
```
pspanja, line 15: We can't remove that as this class depends on it in `getRootLocation()`, and `PagerfantaFindTrait` should not be required for it to work.

pspanja (commented):
What about using an abstract trait method to implement access to `Site`? I'm aware it's bad from DX side to require additional implementation to use the trait, but not sure how else to solve that properly.

pspanja:
> I suggest reducing create*Pager() to only two methods, createSearchPager() and createSearchHitPager(). Inside them we can test given $query to decide what adapter to use.

What about this one?

bundle/Controller/Controller.php
emodric: As I understood @MarioBlazek, one of the points was to use `PagerfantaFindTrait` inside this controller too, in which case this method is not needed any more since it's available inside the trait itself.

UPGRADE.md
emodric: This would be the solution to your question about abstract trait method. A minor BC break, yes, but it would not require an additional implementation to use the trait, it would already be implemented with the trait and called when instatiating the controller. Hence requirement to use the correct parent in service definition.

pspanja: It should be explicitly worded, as far as I understand this is really about requiring `Site` injection through `setSite()` method.

bundle/Controller/Controller.php
pspanja: That's kind of obvious, but what if we introduce another trait instead:

```php
SiteAwareTrait
{
    public setSite($site);
    protected getSite();
}
```

Then we can use it to implement abstract trait method `PagerfantaFindTrait::getSite()`.

What do you think?

MarioBlazek: +1 for `SiteAwareTrait`, that was my first idea to go this way.

emodric: `SiteAwareTrait` is fine, but as I understood, you were against it @pspanja ? :)

MarioBlazek:
@pspanja we actually agreed to have explicit methods.

bundle/Controller/Controller.php
pspanja: I was? Where?

MarioBlazek: We talked in person :)

pspanja: Oh well :D

MarioBlazek:
@emodric @pspanja can we then summarize what needs to be done ?

emodric:
Tha way it is now, is fine by me, with explicit upgrade instructions about specifying the parent for controller services.

MarioBlazek:
I've updated PR, @pspanja please check it out.

UPGRADE.md
```diff
@@ -1,6 +1,17 @@
 eZ Platform Site API upgrade instructions
 =========================================
 
+Upgrading from 2.2 to 2.3
+-------------------------
```
emodric: Needs to be 2.4 now :)

MarioBlazek: That's an eagle eye @emodric, updated :)

emodric (approved):

UPGRADE.md
```diff
@@ -1,6 +1,17 @@
 eZ Platform Site API upgrade instructions
 =========================================
 
+Upgrading from 2.3 to 2.4
+-------------------------
+
+Controllers that extend from `Netgen\Bundle\EzPlatformSiteApiBundle\Controller\Controller` and are registered inside dependency injection container must set `parent` config key within service definition, example:
```
pspanja: This is because of `setContainer` and `setSite` calls, right? In that case it's not really a must, I would mention required calls explicitly, and then suggest that it can be done by setting parent configuration which already does it.

Otherwise this looks good :)

MarioBlazek:
@pspanja final updates done.

pspanja (approved):

pspanja:
🎉 Thanks @MarioBlazek!

PR #54, opened 2017-11-21
Allow installing on eZ kernel 7.0

emodric:
Since eZ kernel 7.0 is now in beta, would be good to have it in 2.3 :)


pspanja:
Thanks @emodric!

PR #56, opened 2017-11-28
Fix building repository after implementing tolerant search service

emodric:
https://github.com/ezsystems/ezpublish-kernel/pull/2142 added a new param to `RepositoryFactory::buildRepository`.

This adds the param as an optional dependency (hence `?`) so it doesn't crash on previous versions if service doesn't exist.

### DIFF
--- a/lib/Resources/config/internal.yml
+++ b/lib/Resources/config/internal.yml
@@ -13,6 +13,7 @@ services:
         arguments:
             - '@ezpublish.api.persistence_handler'
             - '@ezpublish.spi.search.legacy'
+            - '@?ezpublish.search.background_indexer'
         lazy: true
         public: false
 


pspanja:
Will this really work with previous versions, since the third argument is `array $serviceSettings = array()`?

emodric:
This is an argument that is passed to `RepositoryFactory::buildRepository`, which is a factory method for `netgen.ezplatform_site.inner_repository`. It is not passed to the factory constructor itself :)

See here for previous version of the factory method: https://github.com/ezsystems/ezpublish-kernel/blob/6.11/eZ/Bundle/EzPublishCoreBundle/ApiLoader/RepositoryFactory.php#L86

pspanja:
Alright then :))

pspanja (approved):

MarioBlazek (approved):

PR #57, opened 2017-12-07
Add adapters from filter and find services to eZ search service

emodric:
These classes are adapters from Site API filter and find services to SearchService interface
from eZ Publish kernel. The point is being able to replace usage of eZ search service
with Site API filter or find service without touching consuming code.

Ofcourse, we need to convert value objects in search hits from Site API to eZ API value objects.

Methods implemented here do not use `$languageFilter` argument since it is handled automatically
by the filter and find services themselves.

As for `$filterOnUserPermissions`, filter and find services don't support it, so it is simply ignored.

Ran some basic tests with replaced services, and it works fine.

### DIFF
--- a/lib/Resources/config/services.yml
+++ b/lib/Resources/config/services.yml
@@ -18,3 +18,21 @@ services:
 
     netgen.ezplatform_site.site:
         alias: 'netgen.ezplatform_site.core.site'
+
+    # Search adapters for filter and find services to
+    # eZ\Publish\API\Repository\SearchService interface
+    # from eZ Publish
+
+    netgen.ezplatform_site.filter_service.search_adapter:
+        class: Netgen\EzPlatformSiteApi\API\Adapter\FilterServiceAdapter
+        public: false
+        arguments:
+            - "@netgen.ezplatform_site.filter_service"
+            - "@ezpublish.spi.search.legacy"
+
+    netgen.ezplatform_site.find_service.search_adapter:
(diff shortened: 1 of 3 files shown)


MarioBlazek (approved):

lib/Resources/config/services.yml
```diff
@@ -18,3 +18,21 @@ services:
...
+        arguments:
+            - "@netgen.ezplatform_site.filter_service"
+            - "@ezpublish.spi.search.legacy"
+
+    netgen.ezplatform_site.find_service.search_adapter:
+        class: Netgen\EzPlatformSiteApi\API\Adapter\FindServiceAdapter
+        public: false
+        arguments:
+            - "@netgen.ezplatform_site.find_service"
+            - "@ezpublish.spi.search.solr"
```
pspanja: This should really use `ezpublish.spi.search_engine` since it's not a given Solr is configured as search engine.
Looks good otherwise :)

emodric: Hm... Okay. I presume `ezpublish.spi.search.legacy` is fine for filter adapter?

pspanja: Yes, that's hardcoded in the config here.

emodric: Fixed!

pspanja (approved):
Thanks @emodric! 👍

PR #58, opened 2017-12-14
[3.0] Remove usage of class params

emodric:
Not sure if this should land in 2.x due to removal of two class params from config (for URL Alias router and PreviewController), but I'd be for it.


pspanja (approved):

pspanja:
Thanks @emodric :)

PR #60, opened 2017-12-20
[3.0] Implement Twig runtimes and use Twig namespaces

emodric:
* Twig runtimes provide separating Twig extension implementation from their definition
* Twig namespaces are available since newer versions of Twig


emodric:
Need to test on Symfony 2.8, I don't think Twig runtimes are available over there.

MarioBlazek (approved):
Looks good to me

pspanja (approved):
Thanks @emodric!

PR #63, opened 2018-01-16
Enable PHPUnit 6

MarioBlazek:


composer.json
```diff
@@ -14,8 +14,8 @@
         "ezsystems/ezpublish-kernel": "^6.0|^7.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.8.35|^5.7",
-        "matthiasnoback/symfony-dependency-injection-test": "~1.0"
+        "phpunit/phpunit": "^4.8.35|^5.7|^6.0",
```
emodric, line 17: Do we still need PHPUnit 4?

MarioBlazek: we have PHP 5.5

MarioBlazek: maybe it is time to drop support for PHP 5.5 and 5.6, that should be something for 3.0 I suppose ?

emodric: Doesn't have to be. Bumping dependency requirements in minor version is not a breaking change: http://doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html

I wouldn't drop support for PHP 5.6 though, only 5.5, seing how 5.6 is the only officially supported PHP version for eZ Platform 1.x

phpunit-integration-legacy.xml
```diff
@@ -6,6 +6,7 @@
          convertErrorsToExceptions="true"
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
+         beStrictAboutTestsThatDoNotTestAnything="false"
```
emodric: Rather than using this flag, why not refactor `Netgen\EzPlatformSiteApi\Tests\Integration\RelationServiceTest` to replace the offending test with a private method that is not a test itself?

emodric (approved):
Looks good. Dropping PHP versions can be discussed in separate issue.

MarioBlazek:
@emodric https://github.com/netgen/ezplatform-site-api/pull/65

MarioBlazek:
ping @pspanja

pspanja (approved):
Thanks @MarioBlazek!

Issue #61, opened 2018-01-09
Anonymous users do not have access to content.owner property


emodric:
A usual usecase is to display the name of the content object owner, like the blog post author or so on, but using `{{ content.owner.name }}` with anonymous users results in `UnauthorizedException`: `User does not have access to 'read' 'content' with: contentId '14'`

emodric:
Fixed by #66

PR #66, opened 2018-01-16
Provides fix for #61

MarioBlazek:


tests/lib/Unit/Core/Site/Values/ContentTest.php
```diff
@@ -249,10 +224,11 @@ protected function getDomainObjectMapper()
 
         $this->domainObjectMapper = new DomainObjectMapper(
             $this->getSiteMock(),
-            $this->getContentServiceMock(),
-            $this->getContentTypeServiceMock(),
-            $this->getFieldTypeServiceMock(),
-            $this->getUserServiceMock()
+            $this->getRepositoryMock()
+//            $this->getContentServiceMock(),
```
pspanja: You can remove commented code.

lib/Core/Site/Values/Content.php
```diff
@@ -388,11 +395,17 @@ private function getOwner()
...
-            // Do nothing
-        }
+        $loadService = $this->site->getLoadService();
+        $ownerId = $this->getContentInfo()->ownerId;
+        $this->owner = $this->repository->sudo(
+            function() use ($loadService, $ownerId) {
+                try {
+                    return $loadService->loadContent($ownerId);
+                } catch (NotFoundException $e) {
+
```
pspanja: Please add `// Do nothing` comment here, so that the intention is clear.

lib/Core/Site/Values/Content.php
```diff
@@ -388,11 +395,17 @@ private function getOwner()
             return $this->owner;
         }
 
-        try {
-            $this->owner = $this->site->getLoadService()->loadContent($this->getContentInfo()->ownerId);
-        } catch (NotFoundException $e) {
-            // Do nothing
-        }
+        $loadService = $this->site->getLoadService();
```
emodric: You don't have to do this. Closures are automatically bound to the object, so they have access to `$this`.

lib/Core/Site/Values/Content.php
emodric: Please add a comment (`// Do nothing`) indicating that the empty catch is intentional

lib/Core/Site/Values/Content.php
```diff
@@ -408,11 +421,17 @@ private function getInnerOwnerUser()
             return $this->innerOwnerUser;
         }
 
-        try {
-            $this->innerOwnerUser = $this->userService->loadUser($this->getContentInfo()->ownerId);
-        } catch (NotFoundException $e) {
-            // Do nothing
-        }
+        $userService = $this->userService;
```
emodric: There's a difference in allowing anonymous access to user content and user object. While former should be allowed to make it possible to render owner name and so on, this here should not, since it contains sensitive data: username, hashed password and email address, so this needs to revert to the original code.

lib/Core/Site/DomainObjectMapper.php
```diff
@@ -95,6 +92,7 @@ public function mapContent(VersionInfo $versionInfo, $languageCode)
                 'domainObjectMapper' => $this,
                 'contentService' => $this->contentService,
```
emodric: We don't have to inject content service and user service now, right?

emodric (changes requested):

lib/Core/Site/DomainObjectMapper.php
```diff
@@ -51,25 +48,25 @@
...
+     */
+    private $repository;
+
     /**
      * @param \Netgen\EzPlatformSiteApi\API\Site $site
-     * @param \eZ\Publish\API\Repository\ContentService $contentService
-     * @param \eZ\Publish\API\Repository\ContentTypeService $contentTypeService
-     * @param \eZ\Publish\API\Repository\FieldTypeService $fieldTypeService
-     * @param \eZ\Publish\API\Repository\UserService $userService
+     * @param \eZ\Publish\API\Repository\Repository $repository
```
emodric, line 48: This needs update to service definition, right?

MarioBlazek:
@emodric ah, yes, it is already updated in https://github.com/netgen/ezplatform-site-api/pull/63

emodric:
`contentService` and `userService` properties in `DomainObjectMapper.php` seem to be unused now :)

Same goes for `contentTypeService`, `userService` and `fieldTypeService` in `Site.php`

Do verify it :)

emodric:
P.S. This should target 2.3 branch instead of the master, since this is a bugfix.

emodric (approved):

pspanja (approved):

pspanja:
Thanks @MarioBlazek!

PR #67, opened 2018-01-25
Fix setting current user to multiple Repositories

### LINKED ISSUE #64: Inconsistent results
I have started to use this bundle with ezplatform-kernel 7 but it seems filter service don't work properly.

I hace the following controller and results differ from eZ Search Service. Whats wrong in it?
```
<?php

namespace AppBundle\Controller;

use eZ\Publish\API\Repository\Values\Content\Query\Criterion\ParentLocationId;
use Netgen\Bundle\EzPlatformSiteApiBundle\Controller\Controller;
use Netgen\Bundle\EzPlatformSiteApiBundle\View\ContentView;
use eZ\Publish\API\Repository\Values\Content\LocationQuery;


class ContentController extends Controller
{
    /**
     * @param \Netgen\Bundle\EzPlatformSiteApiBundle\View\ContentView $view
     *
     * @return \Netgen\Bundle\EzPlatformSiteApiBundle\View\ContentView
     */
    public function indexAction(ContentView $view)
    {
        $locationId = $view->getParameter('locationId');

        $filterService = $this->getSite()->getFilterService();

        $seachService = $this->get('ezpublish.api.service.search');

        $query = new LocationQuery(['filter' => new ParentLocationId($locationId)]);
        $searchResult = $filterService->filterLocations($query);

        // Total count: 0
        dump($searchResult->totalCount);

        $searchResult = $seachService->findContent($query);

        // Total count: 3
        dump($searchResult->totalCount);
        
        return $view;
    }
}

### PR DESCRIPTION
**Pull request by pspanja**
Fixes #64 

This fixes the problem of setting current Repository user, where Repository instance used to get SearchService with Legacy Search Engine (used for FilterService) was not updated. Current User is the only state Core Repository keeps.

The problem is fixed with new Aggregate Repository implementation, which is registered as public Repository (it rewrites `ezpublish.api.repository` alias). This implementation aggregates top eZ Platform Repository implementation (default `ezpublish.api.repository` alias) and our custom instance and updates current user on all of them.

### DIFF
--- a/lib/Core/Repository/Aggregate/Repository.php
+++ b/lib/Core/Repository/Aggregate/Repository.php
@@ -41,20 +41,7 @@ class Repository implements RepositoryInterface
         array $customRepositories = []
     ) {
         $this->ezRepository = $ezRepository;
-
-        foreach ($customRepositories as $repository) {
-            $this->addRepository($repository);
-        }
-    }
-
-    /**
-     * Add a $repository to the internal collection.
-     *
-     * @param \eZ\Publish\API\Repository\Repository $repository
-     */
-    public function addRepository(RepositoryInterface $repository)
-    {
-        $this->customRepositories[] = $repository;
+        $this->customRepositories = $customRepositories;
     }
 
     public function getCurrentUser()
--- a/lib/Core/Repository/Aggregate/Repository.php
+++ b/lib/Core/Repository/Aggregate/Repository.php
@@ -27,7 +27,7 @@ class Repository implements RepositoryInterface
     /**
      * @var \eZ\Publish\API\Repository\Repository[]
      */
-    private $customRepositories;
+    private $customRepositories = [];
 
     /**
      * Construct repository object from top eZ Platform Repository and
@@ -38,10 +38,23 @@ class Repository implements RepositoryInterface
      */
     public function __construct(
         RepositoryInterface $ezRepository,
-        array $customRepositories
+        array $customRepositories = []
     ) {
         $this->ezRepository = $ezRepository;
-        $this->customRepositories = $customRepositories;
+
+        foreach ($customRepositories as $repository) {
+            $this->addRepository($repository);
+        }
+    }
+
+    /**
+     * Add a $repository to the internal collection.
+     *
+     * @param \eZ\Publish\API\Repository\Repository $repository
+     */
+    public function addRepository(RepositoryInterface $repository)
+    {
+        $this->customRepositories[] = $repository;
     }
 
     public function getCurrentUser()
--- a/lib/Core/Repository/Aggregate/Repository.php
+++ b/lib/Core/Repository/Aggregate/Repository.php
@@ -1,0 +1,180 @@
+<?php
+
+namespace Netgen\EzPlatformSiteApi\Core\Repository\Aggregate;
+
+use eZ\Publish\API\Repository\Repository as RepositoryInterface;
+use eZ\Publish\API\Repository\Values\ValueObject;
+use eZ\Publish\API\Repository\Values\User\UserReference;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by emodric** [comment]
I'll look at it soon.

For now, only a coment that this should probably go against 2.1 branch. That's where filter service was introduced.

**Comment by pspanja** [comment]
Agreed, this is to `master` for testing only.

**Comment by MarioBlazek** [comment]
I can confirm, it does solve issue in my case.

**Comment by pspanja** [comment]
BTW, if someone already implements a custom Repository layer, it would have to be done like here, by replacing `ezpublish.api.repository` alias. Not tested, but in that case this should work transparently with it.

**Comment by emodric** [comment]
I realize that there is no sane way to do this, but just for the record, I'm not sure I like replacing the `ezpublish.api.repository` alias. It feels too much invasive and I think it increases the chances of something breaking later on.

That being said, if we're going to go this way, it would be great if we could add custom repositories to the agreggate one with tags, so potential custom repositories can be added to the aggregate automatically, without having to modify the aggregate repo service definition.

Other than that, this looks good :+1:

**Review by MarioBlazek — APPROVED** [review]

**Comment by pspanja** [comment]
@emodric 

> I realize that there is no sane way to do this, but just for the record, I'm not sure I like replacing the `ezpublish.api.repository` alias. It feels too much invasive and I think it increases the chances of something breaking later on.

Actually, that should be possible, one should be able to implement own API layer and AFAIK this would be the way to do it. The weird part is second instance of Repository, but we had that before. And probably will have it until `ezpublish-kernel` implementation is decoupled so that it becomes possible to build `SearchService` in a sane way.

File: lib/Resources/config/internal.yml [hunk 1]
```diff
@@ -5,6 +5,14 @@ parameters:
     netgen.ezplatform_site.root_location_id: 2
 
 services:
+    #
+    netgen.ezplatform_site.aggregate_repository:
+        class: Netgen\EzPlatformSiteApi\Core\Repository\Aggregate\Repository
+        arguments:
```
emodric: This should be a private service by default.

pspanja: Updated in https://github.com/netgen/ezplatform-site-api/pull/67/commits/a822caa4a2a7259446eb521e155a0c735f705f63.

pspanja:
@emodric implemented custom repo registration through service tag in https://github.com/netgen/ezplatform-site-api/pull/67/commits/f7a8deec58ea04e110ba5bacbf4eb86f76be32b3

@MarioBlazek please test

lib/Core/Repository/Aggregate/Repository.php
```diff
@@ -38,10 +38,23 @@ class Repository implements RepositoryInterface
      */
     public function __construct(
         RepositoryInterface $ezRepository,
-        array $customRepositories
+        array $customRepositories = []
```
emodric: Do we now need a second param here?

pspanja: We don't use it here, I just always add it to aggregate implementation of something.

emodric: Then you don't need `addRepository` method, since you can use `$definition->replaceArgument(1, [...])` in the compiler pass, where `[...]` is an array of `Reference` objects.

pspanja: Agreed, updated in https://github.com/netgen/ezplatform-site-api/pull/67/commits/45911390a2a98b6d74383bf0dd57435aba957503

emodric:
:+1:, with rebase to `2.1-release`

emodric (approved):

pspanja:
Thanks guys, merged manually.

PR #68, opened 2018-02-02
[3.0] Introduce limit parameter to loadFieldRelations method

mivancic:
In order to skip the slicing of the filter query result array which contains all the relations, limit parameter needs to be propagated all the way to the getRelatedContentItems method which will ensure that the resulting array has limited result returned.

This will not affect/break previous versions

### DIFF
--- a/lib/Core/Site/RelationService.php
+++ b/lib/Core/Site/RelationService.php
@@ -87,7 +87,7 @@ class RelationService implements RelationServiceInterface
     }
 
     /**
-     * Return an array of all related Content from the given arguments if $limit is not provided.
+     * Return an array of related Content items, optionally limited by $limit.
      *
      * @param array $relatedContentIds
      * @param array $contentTypeIdentifiers
--- a/lib/Core/Site/RelationService.php
+++ b/lib/Core/Site/RelationService.php
@@ -68,7 +68,7 @@ class RelationService implements RelationServiceInterface
         $contentId,
         string $fieldDefinitionIdentifier,
         array $contentTypeIdentifiers = [],
-        $limit = 0
+        int $limit = 0
     ): array {
         $content = $this->site->getLoadService()->loadContent($contentId);
 
@@ -97,7 +97,7 @@ class RelationService implements RelationServiceInterface
      *
      * @return \eZ\Publish\API\Repository\Values\Content\Content[]
      */
-    private function getRelatedContentItems(array $relatedContentIds, array $contentTypeIdentifiers, $limit = 0): array
+    private function getRelatedContentItems(array $relatedContentIds, array $contentTypeIdentifiers, int $limit = 0): array
     {
         if (count($relatedContentIds) === 0) {
             return [];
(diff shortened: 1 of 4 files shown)


MarioBlazek (approved):

lib/Core/Site/RelationService.php
```diff
@@ -107,7 +110,7 @@ private function getRelatedContentItems(array $relatedContentIds, array $content
 
         $query = new Query([
             'filter' => $criteria,
-            'limit' => count($relatedContentIds),
+            'limit' => $limit ? $limit : count($relatedContentIds),
```
emodric: This should probably be cast to `int` and check if larger than zero before using it.

lib/Core/Site/Values/Content.php
```diff
@@ -308,12 +308,12 @@ public function getFieldRelation($fieldDefinitionIdentifier)
 
     public function getFieldRelations($fieldDefinitionIdentifier, $limit = 25)
     {
-        $relations = $this->site->getRelationService()->loadFieldRelations(
+        return $this->site->getRelationService()->loadFieldRelations(
             $this->id,
             $fieldDefinitionIdentifier
+            [],
```
emodric, line 322: Missing a coma in the line above, after `$fieldDefinitionIdentifier`.

emodric (changes requested):

lib/Core/Site/RelationService.php
mivancic: Thanks. Fixed in [01ae43](https://github.com/netgen/ezplatform-site-api/pull/68/commits/01ae43193b234df92a60bfa5c301525934651584)

lib/Core/Site/Values/Content.php
mivancic: Fixed in [01ae43](https://github.com/netgen/ezplatform-site-api/pull/68/commits/01ae43193b234df92a60bfa5c301525934651584)

emodric (approved):

MarioBlazek:
@pspanja we need your input here :)

pspanja:
The problem here is that order of results we get back from the search engine is not defined, so if we limit in the query, we can't be sure to get the correct slice.

For example, if the related IDs are (sorted) `2, 7, 4, 5, 1` and we limit to 3, expected result would be `2, 7, 4`. But the engine might actually return `1, 5, 4`, so we will end up with `4, 5, 1`. That is the reason I left out the limit when implementing this.

Solution would be to have a special sort clause that can sort by the given order of IDs, but I don't see it's possible to implement that in a sane way. If we can't do that, we will have to load the complete set and limit after sorting. Your feedback is welcome.

pspanja:
> Solution would be to have a special sort clause that can sort by the given order of IDs, but I don't see it's possible to implement that in a sane way. If we can't do that, we will have to load the complete set and limit after sorting.

Another alternative would be taking a slice of relation IDs first, but this also not complete because some of the relations might not get returned, either because of permissions or inconsistent data. I went with the second approach -- loading the complete set and limiting after sorting.

pspanja:
Thanks @mivancic!

PR #69, opened 2018-03-13
Fix building repository on eZ kernel 7.1

emodric:
https://github.com/ezsystems/ezpublish-kernel/pull/2253 added a new parameter to `RepositoryFactory::buildRepository`, so we need to add it here.

Similar to https://github.com/netgen/ezplatform-site-api/pull/56


emodric:
Merging, needed for latest 6.7 and 6.13 too.

PR #70, opened 2018-03-18
Implement support for QueryTypes

pspanja:
This adds support for QueryTypes to the SiteAPI. The implementation differs from the `ezpublish-kernel` implementation so that it allows defining multiple queries on any controller. It is also possible to define "named queries", which can be referenced in query configuration for a particular content view. When doing that, it is possible to override any property from named query configuration. Example:

```yaml
ezpublish:
    system:
        frontend_group:
            ng_named_query:
                reverse_relations_named_query:
                    query_type: 'SiteAPI:Content/Relations/ReverseFields'
                    max_per_page: 10
                    page: '@=queryParam("page", 1)'
                    parameters:
                        relation_field: 'relation_field_identifier'
                        content_type:
                            - 'blog_post'
                        sort:
                            - 'field/blog_post/title asc'
            ngcontent_view:
                full:
                    ng_news:
                        template: "@ezdesign/content/full/ng_news.html.twig"
                        match:
                            Identifier\ContentType: ng_news
                        queries:
                            reverse_relations: reverse_relations_named_query
                            reverse_relations_overridden:
                                named_query: reverse_relations_named_query
                                max_per_page: 5
                            forward_relations:
                                query_type: 'SiteAPI:Content/Relations/ForwardFields'
                                max_per_page: 10
                                page: '@=request.query.get("page", 1)'
                                parameters:
                                    relation_field: 'relation_field_identifier'
                                    publication_date:
                                        gt: '10 October 1998'
                                        lt: 'now -5 days'
                                    field:
                                        title: 'Myrtle the title'
                                    content_type:
                                        not: 'article'
                                    sort: 'published desc'
                        params:
                                ...
```

Note that key under `queries` param is used as query "name". Injected queries will be available in Twig templates by that name, through 'ng_query' and 'ng_raw_query' functions. The queries will actually be executed only when accessed through these functions:

```twig
{% for relation in ng_query( 'forward_relations' ) %}
    ...
{% endfor %}

{% for relation in ng_raw_query( 'reverse_relations' %}
    ...
{% endfor %}
```

Difference between is that `ng_query` will return a `Pagerfanta` instance, while `ng_raw_query` will return `SearchResult`.

### On how QueryTypes are implemented here

This still uses `QueryType` infrastructure from eZ kernel, so your existing implementation should work without (or at least with very little) modification. However this also builds on top of it to provide additional functionality. Base `QueryType` is extended with `supportsParameter($name)` method, used to automatically inject supported variables. A set of QueryTypes is provided out of the box, on top of a comprehensive system of base QueryTypes. The main intention behing it is to provide a base for common and predicable behavior and a way to easily build new functionality upon that.

Short description of the main configuration options:

- `query_type` - name of the QueryType
- `named_query` - name of the configured named QueryType
- `use_filter` - whether to use FilterService instead of FindService (true by default)
- `max_per_page` - max results per page for Pagerfanta (25 by default)
- `page` - current page for Pagerfanta (1 by default)
- `parameters` - QueryType parameters, specific per QueryType

For named queries `query_type` is mandatory. For actualy queries configured under `ngcontent_view`, one of `query_type` and `named_query` is mandatory. All of these exept `parameters` are actually options wrapping actual PHP implementation of the `QueryType` interface, which is therefore not aware of it.

Provided QueryTypes deserve dedicated doucmentation, which is out of scope here. Following is description of some of the common behavior. Most of the parameters are related to search criteria. Most of these accept multiple ways of defining them. For example, it is not necessary to provide an operator, it will be automatically resolved from the provided value. For a single value the behavior is as follows:

```
content_type: article
```

The above is identical to:

```
content_type:
    eq: article
```

You can also define an array of values:

```
content_type: [article, blog_post]
```

This is identical to:

```
content_type:
    in: [article, blog_post]
```

It is also possible to define multiple operators for a same criterion, and you can also use a negation:

```
depth:
    gte: 5
    lt: 12
    not:
        in: [7, 9]
```

Field configuration also requires a target, which is the identifier of the field. Example:

```
field:
    date_field:
        not:
            gt: 'today +5 days'
    price:
        between: [100, 200]
        not: 155
```

A helper service to define sort clause from string is implemented, called `SortClauseParser`. This is to enable defining sort clauses from view configuration and Twig templates. Supported sort clause strings:

- `depth [asc|desc]`
- `field/article/title [asc|desc]`
- `modified [asc|desc]`
- `name [asc|desc]`
- `priority [asc|desc]`
- `published [asc|desc]`

Following QueryTypes implemented:

- `SiteAPI:Content/Fetch` (general purpose Content query)
- `SiteAPI:Location/Fetch` (general purpose Location query)
- `SiteAPI:Content/Relations/AllTagFields` (all Content tag fields relations)
- `SiteAPI:Content/Relations/ForwardFields` (normal/forward Content fields relations)
- `SiteAPI:Content/Relations/ReverseFields` (reverse Content fields relations)
- `SiteAPI:Content/Relations/TagFields` (tag relations by specific Content fields)
- `SiteAPI:Location/Children` (Location children)
- `SiteAPI:Location/Siblings` (Location siblings)
- `SiteAPI:Location/Subtree` (Location subtree)

### TODOs

- [x] discussion
- helper service to allow default parameter value in query configuration, for parameters from the `View` object (for example for initial pagination param values)
- do we really need multiple sort clauses and how to best define them from Twig?
- [x] manual testing
- [x] automated testing
- [x] `ng_query( 'relations', false )` Twig function, second param controls use of Pagerfanta


bundle/EventListener/QueryTypeListener.php
```diff
@@ -0,0 +1,53 @@
...
+     */
+    public function __construct(QueryCollectionMapper $queryCollectionMapper)
+    {
+        $this->queryCollectionMapper = $queryCollectionMapper;
+    }
+
+    public static function getSubscribedEvents()
+    {
+        return [
+            MVCEvents::PRE_CONTENT_VIEW => [
```
emodric: You should listen to `eZ\Publish\Core\MVC\Symfony\View\ViewEvents::FILTER_VIEW_PARAMETERS` event, which exists for this exact purpose.

See here for examples: https://github.com/ezsystems/ezpublish-kernel/tree/master/eZ/Publish/Core/MVC/Symfony/View/ParametersInjector

lib/Resources/config/query_types.yml
```diff
@@ -0,0 +1,24 @@
+parameters:
+
+services:
```
emodric: `public: false` on all services :)

lib/Core/Site/QueryTypes/TagRelations.php
```diff
@@ -0,0 +1,103 @@
...
+use eZ\Publish\API\Repository\Values\Content\Query\Criterion\LogicalAnd;
+use eZ\Publish\Core\QueryType\OptionsResolverBasedQueryType;
+use eZ\Publish\Core\QueryType\QueryType;
+use Netgen\EzPlatformSiteApi\API\Values\Content;
+use Netgen\TagsBundle\API\Repository\Values\Content\Query\Criterion\TagId;
+use Netgen\TagsBundle\API\Repository\Values\Tags\Tag;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+/**
+ * todo
```
emodric: ? :) (on other query types too)

lib/Core/Site/QueryTypes/TagRelations.php
```diff
@@ -0,0 +1,103 @@
...
+        return 'SiteAPI:TagFieldRelations';
+    }
+
+    /**
+     * @inheritdoc
+     *
+     * @throws \Symfony\Component\OptionsResolver\Exception\UndefinedOptionsException
+     * @throws \Symfony\Component\OptionsResolver\Exception\AccessException
+     */
+    protected function configureOptions(OptionsResolver $optionsResolver)
```
emodric: I love `OptionsResolver` :)

composer.json
```diff
@@ -11,7 +11,8 @@
     ],
     "require": {
         "php": "~5.6|~7.0",
-        "ezsystems/ezpublish-kernel": "^6.0|^7.0"
+        "ezsystems/ezpublish-kernel": "^6.0|^7.0",
+        "netgen/tagsbundle": "^3.2"
```
emodric: We should probably check that TagsBundle is active too. Something like: https://github.com/netgen/TagsBundle/blob/3.2-release/bundle/DependencyInjection/NetgenTagsExtension.php#L26-L30

Or thrown an exception in content methods working with tags if the bundle is not active.

I'd rather have second one, since we do not want to force others to use TagsBundle just to use site API.

bundle/Templating/Twig/Extension/SortClauseExtension.php
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * @param \Netgen\Bundle\EzPlatformSiteApiBundle\Search\SortClauseParser
+     */
+    public function __construct(SortClauseParser $sortClauseParser)
+    {
+        $this->sortClauseParser = $sortClauseParser;
+    }
+
+    public function getName()
```
emodric: No need for this one. Twig does not use this since ages.

bundle/Resources/config/services.yml
```diff
@@ -42,8 +42,37 @@ services:
         tags:
             - {name: twig.extension}
 
+    netgen.ezplatform_site.sort_clause_parser:
```
emodric, line 45: `public: false` for all services.

emodric (dismissed):
Nice work overall, left a few nitpicking comments though.

bundle/Resources/config/services.yml
pspanja: Added in https://github.com/netgen/ezplatform-site-api/pull/70/commits/96eea81268c31c3a91bcc77dd39bc9c6a712cbeb

lib/Resources/config/query_types.yml
pspanja: Added in https://github.com/netgen/ezplatform-site-api/pull/70/commits/96eea81268c31c3a91bcc77dd39bc9c6a712cbeb

bundle/Templating/Twig/Extension/SortClauseExtension.php
pspanja: Removed in https://github.com/netgen/ezplatform-site-api/pull/70/commits/9a214e971b5ca0c6e8c28b39d9ea0df87a8a1fad

bundle/EventListener/QueryTypeListener.php
pspanja: Removed in https://github.com/netgen/ezplatform-site-api/pull/70/commits/04f87dc1ccbcaa223eb6826dbcca6e9c829bef78

pspanja:
Refactored the whole configuration and mapping implementation in https://github.com/netgen/ezplatform-site-api/pull/70/commits/04f87dc1ccbcaa223eb6826dbcca6e9c829bef78

Configuration is now moved one level up and is also defined through semantic configuration.
`QueryCollection` is now mapped through our own implementation of `ViewProvider` and listener is removed.

pspanja:
We can now also use Request in configuration language expressions, for example for setting current page to the query. Configuration key and template variable is now changed to `find`. Better suggestions welcome :)

Examples are also updated.

pspanja:
Refactored again in https://github.com/netgen/ezplatform-site-api/pull/70/commits/fd84114c91c664ad61cce0e10b31170ca86f922c.

We now have a Twig function `ng_query`, accepting query name and optionally a boolean value to control usage of pager. Pager parameters `max_per_page` and `page` are moved out of `QueryType`, they did not belong there anyway. They are now part of the semantic configuration and are being mapped to the `QueryDefinition`.

lib/Core/Site/QueryTypes/ForwardFieldRelations.php
```diff
@@ -0,0 +1,116 @@
...
+        $sortClauses = $parameters['sort_clauses'];
+
+        if (!$content->hasField($fieldDefinitionIdentifier)) {
+            throw new RuntimeException(
+                "Content does not contain a field '{$fieldDefinitionIdentifier}'"
+            );
+        }
+
+        $field = $content->getField($fieldDefinitionIdentifier);
+        assert($field instanceof Field);
```
emodric: Behavior of `assert` is dependant on PHP configuration, so it would probably not be wise to use it like this?

Docs say:

> Assertions should not be used for normal runtime operations like input parameter checks. As a rule of thumb your code should always be able to work correctly if assertion checking is not activated.

pspanja: I was trying it out, because PHPStorm is not smart enough to figure out has/get combo, so it warns about possible null value for no reason. I’ll remove it, we’ll eventually implement Null field anyway.

pspanja: Removed in https://github.com/netgen/ezplatform-site-api/pull/70/commits/4bbffd58bbf5ac4bb000de9fdcbe51c2f3c350e2

bundle/QueryType/QueryExecutor.php
```diff
@@ -0,0 +1,125 @@
...
+     * @param \Netgen\Bundle\EzPlatformSiteApiBundle\QueryType\QueryDefinition $queryDefinition
+     * @param bool $usePager
+     *
+     * @return \eZ\Publish\API\Repository\Values\Content\Search\SearchResult|\Pagerfanta\Pagerfanta
+     */
+    private function getLocationResult(LocationQuery $query, QueryDefinition $queryDefinition, $usePager)
+    {
+        if ($usePager) {
+            $pager = new Pagerfanta(
+                new LocationSearchFilterAdapter($query, $this->filterService)
```
emodric: Maybe a good idea would be to allow to use find adapters too.

bundle/QueryType/QueryExecutor.php
```diff
@@ -0,0 +1,125 @@
...
+            );
+
+            $pager->setNormalizeOutOfRangePages(true);
+            $pager->setMaxPerPage($queryDefinition->maxPerPage);
+            $pager->setCurrentPage($queryDefinition->page);
+
+            return $pager;
+        }
+
+        return $this->filterService->filterLocations($query);
```
emodric: Same goes for allowing findService too

bundle/Resources/config/services.yml
```diff
@@ -42,8 +42,41 @@ services:
...
+        public: false
+
+    netgen.ezplatform_site.query_executor:
+        class: Netgen\Bundle\EzPlatformSiteApiBundle\QueryType\QueryExecutor
+        arguments:
+            - '@ezpublish.query_type.registry'
+            - '@netgen.ezplatform_site.filter_service'
+        public: false
+
+    netgen.ezplatform_site.twig.extension.query:
```
emodric: Should be private

bundle/Resources/config/view.yml
```diff
@@ -9,8 +9,10 @@ services:
             - '@ezpublish.view.view_parameters.injector.dispatcher'
 
     netgen.ezplatform_site.ngcontent_view_provider.configured:
-        class: "%ezpublish.view_provider.configured.class%"
-        arguments: ["@netgen.ezplatform_site.ngcontent_view.matcher_factory"]
+        class: Netgen\Bundle\EzPlatformSiteApiBundle\View\Provider\Configured
```
emodric, line 15: We can make this private too, since you're touching it

bundle/Templating/Twig/Extension/QueryExtension.php
```diff
@@ -0,0 +1,76 @@
...
+use Netgen\Bundle\EzPlatformSiteApiBundle\View\ContentView;
+use Twig_Error_Runtime;
+use Twig_Extension;
+use Twig_SimpleFunction;
+
+/**
+ * Twig extension for executing queries from the QueryCollection injected
+ * into the template.
+ */
+class QueryExtension extends Twig_Extension
```
emodric: What's the usecase for running queries from templates?

bundle/QueryType/QueryCollectionMapper.php
```diff
@@ -0,0 +1,134 @@
...
+    public function map(array $configuration, ContentView $view)
+    {
+        $queryCollection = new QueryCollection();
+
+        foreach ($configuration as $name => $queryConfiguration) {
+            $queryDefinition = new QueryDefinition([
+                'name' => $queryConfiguration['query_type'],
+                'parameters' => $this->resolveParameters($queryConfiguration['parameters'], $view),
+                'maxPerPage' => $this->resolveParameter($queryConfiguration['max_per_page'], $view),
+                'page' => $this->resolveParameter($queryConfiguration['page'], $view),
```
emodric: I hope this geta overriden with `page` query parameter feom the request, so we can actually page :)

bundle/View/Provider/Configured.php
```diff
@@ -0,0 +1,108 @@
...
+     * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentType
+     */
+    public function getView(View $view)
+    {
+        if (($configHash = $this->matcherFactory->match($view)) === null) {
+            return null;
+        }
+
+        // This service is dispatched by the configured view class, so this assertion should be safe
+        assert($view instanceof ContentView);
```
emodric: Still a leftover?

lib/Core/Site/Values/Content.php
```diff
@@ -429,4 +440,183 @@ private function getInnerOwnerUser()
 
         return $this->innerOwnerUser;
     }
+
+    public function getReverseFieldRelation(
```
emodric: Isn't this supposed to use query executor from this PR, instead of building everything manually?

tests/lib/Integration/Resources/config/legacy.yml
```diff
@@ -4,3 +4,7 @@ parameters:
         - 'ger-DE'
 
 services:
+    # In ezpublish-kernel this is defined in bundle, however we use it in
+    # core implementation here
+    ezpublish.query_type.registry:
```
emodric: This one needs `public: true`

bundle/Templating/Twig/Extension/QueryExtension.php
pspanja: I'm really looking for a better syntax when accessing queries from template. Calling a function seems more visible/explicit then accessing a variable. Note that this still can work only the QueryCollection that was injected to the template, so it's not for running arbitrary queries.

In any case queries need to be executed from template, unless we pre-execute them and inject the results instead.

bundle/QueryType/QueryCollectionMapper.php
pspanja: Depends on the configuration, it does if it was configured like that :)

For example `page: '@=request.query.get("page")'`. This is just resolving the configured value, which means executing language expression if used.

bundle/View/Provider/Configured.php
pspanja: Whoops, removed in https://github.com/netgen/ezplatform-site-api/pull/70/commits/abac8b50c33b18f5d56c643ba38786a857ed66dd

tests/lib/Integration/Resources/config/legacy.yml
pspanja: It's public by default, or?

bundle/QueryType/QueryExecutor.php
pspanja: Yup, but any idea from where it should be controlled?

lib/Core/Site/Values/Content.php
pspanja: Nope, QueryExecutor is for view configuration only, this works with query types directly.

tests/lib/Integration/Resources/config/legacy.yml
emodric: From Symfony 3.4+, it's private by default. Other `public: false` are to disable deprecation notices on Symfony 3.4.

bundle/QueryType/QueryCollectionMapper.php
emodric: Ah, nice. Then it needs to be documented well :)

bundle/Templating/Twig/Extension/QueryExtension.php
emodric: My question was about the second point. I'd rather see that queries were executed from an event listener and results injected.

bundle/QueryType/QueryExecutor.php
emodric: In view config, a bool flag, `use_filter` maybe?

bundle/Templating/Twig/Extension/QueryExtension.php
pspanja: OK, discussed this f2f - it's mostly a matter of preference, let's see what @MarioBlazek has to say :)

tests/lib/Integration/Resources/config/legacy.yml
pspanja: OK, but are you sure it needs that? This is for tests only, it's the same as in `ezpublish-kernel` and seems to be working fine.

bundle/QueryType/QueryExecutor.php
pspanja: Agreed, with `use_filter: true` by default.

tests/lib/Integration/Resources/config/legacy.yml
emodric: Ah, sorry, didn't see it was for tests only :octocat:

bundle/Resources/config/services.yml
pspanja: Added in https://github.com/netgen/ezplatform-site-api/pull/70/commits/b7460a99476a36610a2b7ef5364732165eb22a05

bundle/QueryType/QueryExecutor.php
emodric: :+1:

bundle/Resources/config/view.yml
pspanja: Added in https://github.com/netgen/ezplatform-site-api/pull/70/commits/6c81200e1bed00a3f85967e1c69744defc3bfd20

bundle/QueryType/QueryExecutor.php
pspanja: Added in https://github.com/netgen/ezplatform-site-api/pull/70/commits/bb217e92dd679755ea5074eb5acb2567ef4c74b8

bundle/QueryType/QueryExecutor.php
pspanja: Added in https://github.com/netgen/ezplatform-site-api/pull/70/commits/bb217e92dd679755ea5074eb5acb2567ef4c74b8

composer.json
pspanja: Ok, then `netgen/tagsbundle` goes to the suggest section. Any input on the minimal version we should support?

emodric: `^3.0` should be okay.

emodric: You can use `\Netgen\TagsBundle\Version::VERSION_ID >= 30000` condition.

pspanja:
OK, here comes another twist :)

In https://github.com/netgen/ezplatform-site-api/pull/70/commits/9621d507f87a1f6bb1490ee810d139a15ecda70e I removed all newly added methods from `Content` object. These are all still available through query type / content view configuration, so having them was really only a convenience. My reasoning is: this is still "content model logic", but since it will not be needed very often it's not essential to have it on the `Content`, view config or query types from PHP should suffice. Also, those methods were a bit unwieldy because of all the parameters. Thoughts?

Together with this, if I find the way to handle forward relations from query type properly (we need to load them all to sort them), we could deprecate `RelationService` and remaining relation methods from `Content` as well.

emodric:
:+1: for removal of methods from `Content`

MarioBlazek:
Just tested, works nice :+1:

MarioBlazek:
As an idea in future improvements, it would be very useful to have somehow `reusable` query configuration, something like this:

```yaml
ezpublish:
    system:
        frontend_group:
            ngqueries:
                members:
                    query_type: 'RazaTehnoparkBundle:Search'
                    max_per_page: 12
                    page: '@=request.query.get("page", 1)'
                    parameters:
                        location: '@=location'
                        fetch_subtree: '@=content.getField("page_limit").value'
                        content_type: '@=request.query.get("type", "tpz_patent")'
                        search_text: '@=request.query.get("searchText", "")'

            ngcontent_view:
                full:
                    tpz_members:
                        template: "@ezdesign/content/full/tpz_members.html.twig"
                        match:
                            Id\Location: '%tpz.default.location.members%'
                        queries:
                            - members
                    tpz_external_members:
                        template: "@ezdesign/content/full/tpz_members.html.twig"
                        match:
                            Id\Location: '%tpz.default.location.external_members%'
                        queries:
                            - members
                    tpz_teams:
                        template: "@ezdesign/content/full/tpz_members.html.twig"
                        match:
                            Id\Location: '%tpz.default.location.teams%'
                        queries:
                            - members
                    tpz_alumni:
                        template: "@ezdesign/content/full/tpz_members.html.twig"
                        match:
                            Id\Location: '%tpz.default.location.alumni%'
                        queries:
                            - members
```

MarioBlazek (dismissed):

pspanja:
Implemented support for common/named query configuration. Description is updated to show how it's used. Also note new `queryParam()` function, which is a shortcut to `request.query.get()`.

pspanja:
I think this looks quite good in general, now looking into how to simplify and standardize query type configuration. Some ideas:

- automatically inject Content and Location
- simplify param names
  - field_definition_identifier => field
  - content_type_identifiers => types
  - sort_clauses => sort
- allow param single and multiple values (eg. a string and an array of strings)
- allow passing sort clause string and parse it in the query type
- allow passing Content/Location or Content/Location ID (what about remote ID?)

So we could go from this:

```yaml
ezpublish:
    system:
        frontend_group:
            ng_queries:
                reverse_relations_named_query:
                    query_type: 'SiteAPI:ReverseFieldRelations'
                    max_per_page: 10
                    page: '@=request.query.get("page", 1)'
                    parameters:
                        content: '@=content'
                        field_definition_identifier: 'relation_field'
                        content_type_identifiers:
                            - 'blog_post'
                        sort_clauses:
                            - '@=sort.parse(field/blog_post/title asc)'
```

To this:

```yaml
ezpublish:
    system:
        frontend_group:
            ng_queries:
                reverse_relations_named_query:
                    query_type: 'SiteAPI:ReverseFieldRelations'
                    max_per_page: 10
                    page: '@=queryParam("page", 1)'
                    parameters:
                        field: 'relation_field'
                        types: 'blog_post'
                        sort: 'field/blog_post/title asc'
```

bundle/Templating/Twig/Extension/QueryExtension.php
pspanja: This now also has second parameter `$usePager = true`, so devs can decide about it from where they need it.

pspanja:
Now updated with support for parameter override both through config (override of the referenced named query) and though Twig, by passing an optional hash of parameters. See examples in the description.

pspanja:
This is now complete and ready for another review.

From the last time, I dropped the possibility to override `QueryType` parameters from Twig. Also we now have a set of out-of-the-box query types built upon base query types which provide a way to easily implement new ones while retaining the base options and behavior. See description for more details.

pspanja:
BTW, two obviously missing QueryType options are Section identifier and ObjectState identifier. Criteria for these are missing in API at the moment (only ID versions are available). Since I don't want to work with IDs on this level, these will be implemented separately and supported later on.

lib/Core/Site/QueryType/Content/Relations/ForwardFields.php
```diff
@@ -0,0 +1,102 @@
...
+     */
+    protected function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver->setRequired([
+            'content',
+            'relation_field',
+        ]);
+
+        $resolver->setAllowedTypes('content', SiteContent::class);
+        $resolver->setAllowedTypes('relation_field', ['string', 'string[]']);
```
emodric: `string[]` won't work on Symfony 2.8. You can use options resolver normalizer to validate if the array has only string values to be backwards compatible with Symfony 2.8 and still retain strict checks.

bundle/Resources/config/services.yml
```diff
@@ -42,8 +42,42 @@ services:
...
+        arguments:
+            - '@request_stack'
+        public: false
+
+    netgen.ezplatform_site.query_type.query_definition_mapper:
+        class: Netgen\Bundle\EzPlatformSiteApiBundle\QueryType\QueryDefinitionMapper
+        arguments:
+            - '@ezpublish.query_type.registry'
+            - '@netgen.ezplatform_site.query_type.parameter_processor'
+            - '$ng_named_query$'
```
emodric: These should be injected with setter injection to avoid reinitilizing services when the scope changes. See docs for more details: https://doc.ezplatform.com/en/2.1/guide/configuration/#using-setter-injection-preferred

bundle/View/ContentView.php
```diff
@@ -17,6 +17,13 @@
  */
 class ContentView extends BaseView implements View, ContentValueView, LocationValueView, EmbedView, CachableView
 {
+    /**
+     * Name of the QueryDefinitionCollection variable injected to the template.
+     *
+     * @see \Netgen\Bundle\EzPlatformSiteApiBundle\QueryType\QueryDefinitionCollection
+     */
+    const QUERY_DEFINITION_COLLECTION_NAME = 'queryDefinitionCollection';
```
emodric: Should be `query_definition_collection`, as per Twig coding standards.

lib/Core/Site/QueryType/Location/Siblings.php
```diff
@@ -0,0 +1,74 @@
...
+        /** @var \Netgen\EzPlatformSiteApi\API\Values\Location $location */
+        $location = $parameters['location'];
+
+        return [
+            new ParentLocationId($location->parentLocationId),
+            new LogicalNot(new LocationId($location->id)),
+        ];
+    }
+
+    protected function getQueryCriteria(array $parameters)
```
emodric: Could we remove these boilerplate empty methods by providing default implementations in abstract query type?

bundle/DependencyInjection/Configuration/Parser/ContentView.php
```diff
@@ -2,10 +2,166 @@
...
+    const QUERY_KEY = 'queries';
     const NODE_KEY = 'ngcontent_view';
     const INFO = 'Template selection settings when displaying a content with Netgen Site API';
+
+    /**
+     * Adds semantic configuration definition.
+     *
+     * @param \Symfony\Component\Config\Definition\Builder\NodeBuilder $nodeBuilder Node just under ezpublish.system.<siteaccess>
+     */
+    public function addSemanticConfig(NodeBuilder $nodeBuilder)
```
emodric, line 21: Is there no other way to inject `queries` config without redefining the whole config tree?

lib/Core/Site/QueryType/Content/Relations/AllTagFields.php
```diff
@@ -0,0 +1,93 @@
+<?php
```
emodric, line 1: These should be registered only if Tags Bundle installed and active. You can use checks similar to this to optionally load yml files with services config: https://github.com/netgen/TagsBundle/blob/3.3-release/bundle/DependencyInjection/NetgenTagsExtension.php#L68-L71

bundle/Resources/config/services.yml
pspanja: Fixed in https://github.com/netgen/ezplatform-site-api/pull/70/commits/9b8e22c6ff3514bad6adbe2af46d368cd20c73ef

bundle/View/ContentView.php
pspanja: Changed to `ng_query_definition_collection` in https://github.com/netgen/ezplatform-site-api/pull/70/commits/48800d146b40405391ed54e2fc6410ec42bad4c2

bundle/DependencyInjection/Configuration/Parser/ContentView.php
pspanja: It's in the middle of it... I don't see a way to inject it into the structure that is already built.

emodric: Yep, too bad :(

lib/Core/Site/QueryType/Location/Siblings.php
pspanja: Yes, updated in https://github.com/netgen/ezplatform-site-api/pull/70/commits/dcf3f8a047522a8592419b36134e392a86706174

lib/Core/Site/QueryType/Content/Relations/AllTagFields.php
pspanja: Added in https://github.com/netgen/ezplatform-site-api/pull/70/commits/e0cc48ff75986bb8d68e505b336344b32f392a11

lib/Core/Site/QueryType/Base.php
```diff
@@ -0,0 +1,428 @@
...
+     * Return query criteria.
+     *
+     * Here you can return null or a Criterion instance.
+     * Override this method as needed.
+     *
+     * @param array $parameters
+     *
+     * @return null|Criterion
+     */
+    protected function getQueryCriteria(array $parameters)
```
RandyCupic: Why is the method called getQueryCriteri**a** (plural) if it returns only single Criterion or null?

lib/Core/Site/QueryType/Content/Relations/AllTagFields.php
RandyCupic: This query type will return all content which contains the corresponding tags, including the current (context) content. Since this represents a relation based on the current content, I think that there's no need to return the current content itself.

For example, when you have an article and you want to fetch related articles based on tags. There's no need to get the current article as a related one as well.

PS. same goes for TagFields query type as well.

lib/Core/Site/QueryType/Content/Relations/ForwardFields.php
pspanja: Updated using `setAllowedValues()` with closures in https://github.com/netgen/ezplatform-site-api/pull/70/commits/e23dd5042ed9eedc1f81b54cb10b92de90e774cd.

This also exposed some mistakes with test fixtures, which were also fixed.
This approach can also be used to make more precise option validation than we do now, but I think we can go with this for beta, and aim to implement better validation for stable or even 2.6.

lib/Core/Site/QueryType/Content/Relations/AllTagFields.php
pspanja: Added in https://github.com/netgen/ezplatform-site-api/pull/70/commits/11eb4a24d755a6850e244f00c5c9bd0b8f32ea15.

I also added `exclude_context` option to control that behavior, defaulting to `true`.
Not sure if the name is good enough, I'm open for better suggestions.

Will also update `Subtree` with `exclude_root = true` instead of `include_root = false`.

lib/Core/Site/QueryType/Base.php
pspanja: Right, updated in https://github.com/netgen/ezplatform-site-api/pull/70/commits/841871b4bff40fdfd0298b8b1a3ccefc51d2b905

pspanja:
Anyways, I changed `exclude_root` and `exclude_context` to `exclude_self` in https://github.com/netgen/ezplatform-site-api/pull/70/commits/4b2496a6b0cd30dd390ef398c12247de79dd1708.

I think I covered all remarks now.

lib/Core/Site/QueryType/CriterionDefinitionResolver.php
```diff
@@ -0,0 +1,208 @@
...
+            if (array_key_exists($key, self::$operatorMap) || 'not' === $key) {
+                $isOperatorMap = true;
+            } else {
+                $isValueCollection = true;
+            }
+        }
+
+        if ($isOperatorMap && $isValueCollection) {
+            throw new InvalidArgumentException(
+                'Array of parameters is not an operator map nor a value collection'
```
RandyCupic: Maybe the exception message is a little bit misleading? :)

pspanja: It is :)

Improved a little bit in https://github.com/netgen/ezplatform-site-api/pull/70/commits/ce90b97c475b5d008d4481a88dfe228184509b70

lib/Core/Site/QueryType/Location/Fetch.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Netgen\EzPlatformSiteApi\Core\Site\QueryType\Location;
+
+use Netgen\EzPlatformSiteApi\Core\Site\QueryType\Location;
+use Symfony\Component\OptionsResolver\OptionsResolver;
```
RandyCupic: Not needed anymore.

emodric (approved):

emodric (approved):

lib/Core/Site/QueryType/Location/Fetch.php
pspanja: Removed in https://github.com/netgen/ezplatform-site-api/pull/70/commits/fcb80b0326fedcfeb5f6a50e610b4fae02ba4385

RandyCupic (approved):

pspanja:
Thanks @RandyCupic, @emodric and @MarioBlazek 🙇

PR #74, opened 2018-04-06
Add searchResult to pagerfanta adapter

wizhippo:
This adds the ability to get the searchResult from the pager

- [x] Add to location adapter too

### DIFF
--- a/lib/Core/Site/Pagination/Pagerfanta/ContentSearchHitAdapter.php
+++ b/lib/Core/Site/Pagination/Pagerfanta/ContentSearchHitAdapter.php
@@ -27,6 +27,11 @@ class ContentSearchHitAdapter implements AdapterInterface
      */
     private $nbResults;
 
+    /**
+     * @var \eZ\Publish\API\Repository\Values\Content\Search\SearchResult
+     */
+    private $searchResult;
+
     public function __construct(Query $query, FindService $findService)
     {
         $this->query = $query;
@@ -65,13 +70,25 @@ class ContentSearchHitAdapter implements AdapterInterface
         $query->limit = $length;
         $query->performCount = false;
 
-        $searchResult = $this->findService->findContent($query);
+        $this->searchResult = $this->findService->findContent($query);
 
         // Set count for further use if returned by search engine despite !performCount (Solr, ES)
-        if (!isset($this->nbResults) && isset($searchResult->totalCount)) {
-            $this->nbResults = $searchResult->totalCount;
+        if (!isset($this->nbResults) && isset($this->searchResult->totalCount)) {
+            $this->nbResults = $this->searchResult->totalCount;
         }
 
-        return $searchResult->searchHits;
+        return $this->searchResult->searchHits;
+    }
+
+    /**
+     * @return \eZ\Publish\API\Repository\Values\Content\Search\SearchResult
+     */
+    public function getSearchResult()
+    {
+        if ($this->searchResult === null) {
+            $this->searchResult = $this->findService->findContent($this->query);
+        }
+
+        return $this->searchResult;
     }
 }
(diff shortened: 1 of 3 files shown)


lib/Core/Site/Pagination/Pagerfanta/ContentSearchHitAdapter.php
```diff
@@ -40,14 +40,13 @@ public function __construct(Query $query, FindService $findService)
      */
     public function getNbResults()
     {
-        if (isset($this->nbResults)) {
-            return $this->nbResults;
+        if ($this->isSearchResultNotCached()) {
+            $countQuery = clone $this->query;
```
wizhippo: Do we really need to clone?

pspanja: Yes, otherwise injected query would be modified by `getNbResults()` and `getSlice()` methods.

pspanja:
Thanks @wizhippo! Is this ready for review now?

wizhippo:
@pspanja Should be good for review now.

pspanja:
How this behaves if `getSearchResult()` is accessed before the iterator? It seems in that case the query will be executed with the initial limit and offset, which can be different than what is set through Pagerfanta.

I think in this case I would prefer that we return null or even throw an exception.

Ping @emodric, @MarioBlazek, @hknezevic, any POV on that?

wizhippo:
One would have to make sure they call `$pagerfanta->getCurrentPageResults();` prior to using the pager in a twig template if they are trying to get the facets before showing the pager.  Otherwise an exception would be could be thrown if they try to get the pagerfanta.adapter.searchResult prior.

pspanja:
Yes. It should not matter for facets, but this can be misused, which means it will be misused :)

emodric:
If we cannot call needed methods from inside the adapter (and I'm not saying we can or can't, it's just a thought), I vote for an exception, it would prevent misuse, and more importantly, probably detect issues in user code.

MarioBlazek:
If we go with this, specific exception must be thrown to point developer about Pager misuse. Once I've called methods in wrong order on Pager, it required some time to discover the problem.

pspanja:
I discussed this a bit with @hknezevic, a better approach would be to expose the facets directly. This goes into the same category with total count, it's about properties that do not depend on limit and offset. Then we do not need to access iterator first, but can execute the query to cache all the properties instead, same way total count is now handled.

In future we can also add max score. What do you think?

emodric:
@pspanja Seems good! :+1:

lib/Core/Site/Pagination/Pagerfanta/ContentSearchHitAdapter.php
```diff
@@ -49,10 +54,21 @@ public function getNbResults()
...
-        $countQuery->limit = 0;
+        return $this->nbResults = $this->getSearchResult()->totalCount;
+    }
+
+    /**
+     * Returns the facets of the results.
+     *
+     * @return \eZ\Publish\API\Repository\Values\Content\Search\Facet[] The facets of the results
+     */
+    public function getFacets()
```
pspanja, line 65: Looks good, but there is no reason for having separate methods for facets and total count. They can both be set from a single search result, not doing that means executing extra query. So maybe something like:

```php
private $areLimitZeroPropertiesInitialized = false;

private function initializeLimitZeroProperties()
{
    // Perform search with limit 0

    $this->facets = ...
    $this->totalCount = ...

    $this->areLimitZeroPropertiesInitialized = true;
}

public function getFacets()
{
    if (!$areLimitZeroPropertiesInitialized) {
        $this->initializeLimitZeroProperties();
    }

    return $this->facets;
}

...
```

wizhippo: The` searchResult` is cached.  A new search is only performed if `$this->searchResult === null`.  `$this->searchResult` will be populated by whichever method performs the query first so no extra query would be performed.  If you prefer your example that is fine too, I can change it.

pspanja: No need, that's good :) Sorry, I didn't look carefully enough.

pspanja (approved):
Looks good, thanks @wizhippo!

PR #76, opened 2018-05-30
Refactor Pagerfanta adapters

pspanja:
This implements new Pagerfanta adapters and deprecates existing ones.

With this change, when deprecations are removed we will have only two adapters, `FilterAdapter` and `FindAdapter`, using `FilterService` and `Findservice`. Location and Content search/filtering is dispatched through checking the type of the given query. New trait `PagerfantaTrait` is added, providing convenience methods `getFilterPager()` and `getFindPager()`. Existing trait `PagerfantaFindTrait` is deprecated.

Adapters return new `Slice` instance, which implements `IteratorAggregate` with `ArrayIterator` (same as returned by Pagerfanta) and also provides method `getSearchHits()`. With that, separate adapters for iterating over SearchHits and value objects contained in them are no longer needed.

They implement new `SearchResultExtras` interface, providing access to facets, max score and execution time.

Adapters are put in use for QueryTypes, in the `QueryExecutor` service.

All added deprecations are targeted for removal in `3.0`.

### DIFF
--- a/lib/Core/Site/Pagination/Pagerfanta/ContentInfoSearchHitAdapter.php
+++ b/lib/Core/Site/Pagination/Pagerfanta/ContentInfoSearchHitAdapter.php
@@ -31,6 +31,11 @@ class ContentInfoSearchHitAdapter implements AdapterInterface
 
     public function __construct(Query $query, FindService $findService)
     {
+        @trigger_error(
+            'ContentInfoSearchHitAdapter is deprecated since version 2.2 and will be removed in 3.0. Use FindAdapter or FilterAdapter instead.',
+            E_USER_DEPRECATED
+        );
+
         $this->query = $query;
         $this->findService = $findService;
     }
--- a/lib/Core/Site/Pagination/Pagerfanta/Slice.php
+++ b/lib/Core/Site/Pagination/Pagerfanta/Slice.php
@@ -1,0 +1,37 @@
+<?php
+
+namespace Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta;
+
+use ArrayIterator;
+use eZ\Publish\API\Repository\Values\Content\Search\SearchHit;
+
+/**
+ * Implements ArrayIterator over SearchHit values and access to the array of
+ * the given SearchHit instances.
+ */
+final class Slice extends ArrayIterator
+{
+    /**
+     * @var \eZ\Publish\API\Repository\Values\Content\Search\SearchHit[]
+     */
+    private $searchHits;
+
+    public function __construct(array $searchHits)
+    {
+        $this->searchHits = $searchHits;
+
+        parent::__construct(
+            array_map(
+                function (SearchHit $searchHit) {
+                    return $searchHit->valueObject;
+                },
+                $searchHits
+            )
+        );
+    }
+
+    public function getSearchHits()
+    {
+        return $this->searchHits;
+    }
+}
--- a/lib/Core/Site/Pagination/Pagerfanta/Slice.php
+++ b/lib/Core/Site/Pagination/Pagerfanta/Slice.php
@@ -4,12 +4,13 @@ namespace Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta;
 
 use ArrayIterator;
 use eZ\Publish\API\Repository\Values\Content\Search\SearchHit;
+use IteratorAggregate;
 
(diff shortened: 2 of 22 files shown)


pspanja:
@wizhippo for some reason I can't add you as a reviewer here, but if you have time your input would be appreciated.

wizhippo (commented):
Much better.  Felt wrong before having so much duplicated in the adapters.

MarioBlazek:
Looks good 👍

lib/Core/Site/Pagination/Pagerfanta/ContentInfoSearchHitAdapter.php
```diff
@@ -31,6 +31,11 @@ class ContentInfoSearchHitAdapter implements AdapterInterface
 
     public function __construct(Query $query, FindService $findService)
     {
+        @trigger_error(
+            'ContentInfoSearchHitAdapter is deprecated since version 2.2 and will be removed in 3.0. Use FindAdapter or FilterAdapter instead.',
```
emodric, line 35: I guess it's `deprecated since version 2.5` :)

lib/Core/Site/Pagination/Pagerfanta/Slice.php
```diff
@@ -0,0 +1,37 @@
...
+namespace Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta;
+
+use ArrayIterator;
+use eZ\Publish\API\Repository\Values\Content\Search\SearchHit;
+
+/**
+ * Implements ArrayIterator over SearchHit values and access to the array of
+ * the given SearchHit instances.
+ */
+final class Slice extends ArrayIterator
```
emodric: Any specific reason why you extended `ArrayIterator`, rather than simply using `IteratorAggregate` interface which returns the `ArrayIterator` instance?

lib/Core/Traits/PagerfantaTrait.php
```diff
@@ -0,0 +1,59 @@
...
+
+use eZ\Publish\API\Repository\Values\Content\Query;
+use Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta\FilterAdapter;
+use Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta\FindAdapter;
+use Pagerfanta\Pagerfanta;
+
+/**
+ * Provides methods to build Pagerfanta instance using on FilterAdapter or FindAdapter.
+ *
+ * Note: expects that SiteAwareTrait is also used.
```
emodric, line 13: Traits can use other traits, so you can just `use` `SiteAwareTrait` here. The comment can still remain, saying something in line of "`Site` object needs to be injected due to usage of `SiteAwareTrait`" or something similar.

pspanja: It would clash with `PagerfantaFindTrait` which also uses it, that's the only reason I left it out. We can add it once that's removed in `3.0`.

emodric: :+1:

lib/Core/Site/Pagination/Pagerfanta/Slice.php
pspanja: I would still need to extract value objects to make them available, so not sure what would be the benefit with that approach?

emodric: I know you would, this way just feels dirty, nothing else. Personal preference, I guess.

lib/Core/Site/Pagination/Pagerfanta/ContentInfoSearchHitAdapter.php
pspanja: Actually we deprecated it in `2.2` (see class docblock) but didn't add the error.

lib/Core/Site/Pagination/Pagerfanta/Slice.php
pspanja: OK, I understand your point, will update.

pspanja: Updated in https://github.com/netgen/ezplatform-site-api/pull/76/commits/69040f20232de29ce00ed8eb25a834d8ba1271e6.

pspanja:
Merged, thanks everyone!

PR #78, opened 2018-06-18
[2.5] Support for Section and ObjectState identifiers with QueryTypes

pspanja:
This implements support for Section and ObjectState identifiers with QueryTypes, through https://github.com/netgen/ezplatform-search-extra (still unreleased).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,8 @@
     ],
     "require": {
         "php": "~5.6|~7.0",
-        "ezsystems/ezpublish-kernel": "^6.0|^7.0"
+        "ezsystems/ezpublish-kernel": "^6.0|^7.0",
+        "netgen/ezplatform-search-extra": "^1.0@dev"
     },
     "require-dev": {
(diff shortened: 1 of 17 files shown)


composer.json
```diff
@@ -11,7 +11,8 @@
     ],
     "require": {
         "php": "~5.6|~7.0",
-        "ezsystems/ezpublish-kernel": "^6.0|^7.0"
+        "ezsystems/ezpublish-kernel": "^6.0|^7.0",
+        "netgen/ezplatform-search-extra": "^1.0-dev"
```
emodric: I think the standard notation is `^1.0@dev`

emodric (approved):
One minor comment, otherwise looks good :+1:

MarioBlazek (approved):
Looks fine :thumbsup:

composer.json
pspanja: Yup, amended the commit.

PR #79, opened 2018-07-26
NSA-3: Fix the setting of nbResults value in the pagerfanta adapters

mivancic:
`getSlice` method in adapters overrides the pager search result with the query parameter `!performCount` and this causes pager to not work properly.

### DIFF
--- a/lib/Core/Site/Pagination/Pagerfanta/ContentSearchHitAdapter.php
+++ b/lib/Core/Site/Pagination/Pagerfanta/ContentSearchHitAdapter.php
@@ -37,7 +37,7 @@ class ContentSearchHitAdapter implements AdapterInterface
     /**
      * @var \eZ\Publish\API\Repository\Values\Content\Search\SearchResult
      */
-    private $searchResult;
+    private $searchResultWithLimitZero;
 
     public function __construct(Query $query, FindService $findService)
     {
@@ -61,7 +61,7 @@ class ContentSearchHitAdapter implements AdapterInterface
             return $this->nbResults;
         }
 
-        return $this->nbResults = $this->getSearchResult()->totalCount;
+        return $this->nbResults = $this->getSearchResultWithLimitZero()->totalCount;
     }
 
     /**
@@ -75,7 +75,7 @@ class ContentSearchHitAdapter implements AdapterInterface
             return $this->facets;
         }
 
-        return $this->facets = $this->getSearchResult()->facets;
+        return $this->facets = $this->getSearchResultWithLimitZero()->facets;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


pspanja:
This will need an update to unit tests as well.

lib/Core/Site/Pagination/Pagerfanta/ContentSearchHitAdapter.php
```diff
@@ -93,18 +93,13 @@ public function getSlice($offset, $length)
         $query->limit = $length;
         $query->performCount = false;
 
-        $this->searchResult = $this->findService->findContent($query);
+        $searchResult = $this->findService->findContent($query);
 
-        // Set count for further use if returned by search engine despite !performCount (Solr, ES)
-        if (!isset($this->nbResults) && isset($this->searchResult->totalCount)) {
```
pspanja: I think the problem was in setting the search result to the property, if that is removed shouldn't it be possible to keep this?

mivancic: I think that we need to remove setting the `facets` property from this method too. Because this method only needs to take care of getting the slice of search result. 

Pagerfanta view render is the one that is responsible for initializing the [Pagerfanta properties](https://github.com/whiteoctober/Pagerfanta/blob/master/src/Pagerfanta/View/DefaultView.php#L49) so we only need to take care of providing the [nbResults](https://github.com/whiteoctober/Pagerfanta/blob/master/src/Pagerfanta/Pagerfanta.php#L361) in our [adapter](https://github.com/netgen/ezplatform-site-api/blob/master/lib/Core/Site/Pagination/Pagerfanta/ContentSearchHitAdapter.php#L58) which is already done.

hknezevic: ~~@mivancic I think setting the facets was added for the convenience, so you don't have to fetch the entire results page to retrieve facets. For instance, instead of doing~~
$pager->getCurrentPageResults();
$facets = $pager->adapter->getFacets();

~~, you can simply call a count method without having to retrieve the results, and have facets provided to you.
But I am not 100% sure. @pspanja WTYT?~~

EDIT: ah, sorry, this is the getSlice() method we're discussing, ignore the previous part of comment. But still, I believe the same reasoning was behind setting facets in this method. Same as the totalCount, the facets have global value and do not depend on the current result. And facets can be retrieved only if the search adapter has already called the getSearchResults. So in this case you can do exactly as described above: retrieve the current page results and at the same time have the facets retrieved for you. Which is kind of convenient.

pspanja: Yes, but but still you shouldn't have to call `$pager->getCurrentPageResults()` just to later get facets from `$pager->adapter->getFacets()`, that should work standalone.

AFAIR the problem was caching the result here in `$this->searchResult`, and this with `$query->performCount = false`. So if you accessed the result first, `nbResults` would be `null`, but if you accessed `nbResults` first, everything was in order. That's why I think it should be possible to keep setting `ngResults` and `facets` here -- since we are not setting `$this->searchResult` any longer.

hknezevic: @pspanja fair point. I think we had an issue or a pull request opened earlier because of the same reason, right?

pspanja: > Same as the totalCount, the facets have global value and do not depend on the current result.

Yes, exactly. It was probably a mistake to include this in the release, deprecating it at the same moment, providing new adapters. But oh well :)

pspanja: @hknezevic hum, not sure, I can't find one.

pspanja (approved):

pspanja:
Great, thanks @mivancic 👍

PR #80, opened 2018-07-26
Use new FilterAdapter instead of deprecated LocationSearchFilterAdapter

mivancic:
Replaced the `LocationSearchFilterAdapter` with the new `FilterAdapter` in following Value objects:

- Content
- ContentInfo
- Location
- Node


pspanja (approved):

pspanja:
Thanks @mivancic!

emodric:
This is a breaking change and should not have been done in a bugfix release (it should go to 3.0).

In 2.5.2, the following code works fine:

```
{% set children = location.filterChildren(['image'], 1).currentPageResults %}
{{ dump(children[0].content.fields.name) }}
```

While in 2.5.3, the exception is thrown:

```
Impossible to access a key "0" on an object of class
"Netgen\EzPlatformSiteApi\Core\Site\Pagination\Pagerfanta\Slice"
that does not implement ArrayAccess interface.
```

pspanja:
Thanks for reporting @emodric, I'll release a fix in few minutes (see #82).

PR #87, opened 2018-08-18
Deprecate from ContentInfo

pspanja:
This deprecates the following from `ContentInfo`:

- method `getLocations()`
- method `filterLocations()`
- property `$content`

The reason is we kept `ContentInfo` only to make migration of vanilla templates easier. So it's really only a container of properties and can't be obtained otherwise than through aggregation in `Content` object (methods providing that exist, but are already deprecated). Also, deprecated methods already exist on `Content` object.

For those reasons `ContentInfo` should not provide the deprecated methods and properties. Properties `$mainLocation` and `$innerContentType` are kept as an exception, since properties `$mainLocationId` and `$contentTypeId` are already available there (and on Repository `ContentInfo`).

### DIFF
--- a/lib/Core/Site/Values/ContentInfo.php
+++ b/lib/Core/Site/Values/ContentInfo.php
@@ -153,11 +153,15 @@ final class ContentInfo extends APIContentInfo
 
     public function getLocations($limit = 25)
     {
+        @trigger_error('getContent() is deprecated since version 2.6 and will be removed in 3.0. Use the same method on Content object instead.', E_USER_DEPRECATED);
+
         return $this->filterLocations($limit)->getIterator();
     }
 
     public function filterLocations($maxPerPage = 25, $currentPage = 1)
     {
+        @trigger_error('getContent() is deprecated since version 2.6 and will be removed in 3.0. Use the same method on Content object instead.', E_USER_DEPRECATED);
+
         $pager = new Pagerfanta(
             new FilterAdapter(
                 new LocationQuery([
--- a/lib/Core/Site/Values/ContentInfo.php
+++ b/lib/Core/Site/Values/ContentInfo.php
@@ -188,6 +188,8 @@ final class ContentInfo extends APIContentInfo
 
     private function getContent()
     {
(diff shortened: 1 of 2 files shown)


pspanja:
I already had this in `master`, but then realized there should probably be a PR and a review :)

MarioBlazek:
Same methods exist on `Content` already, I wouldn't be missing these from `ContentInfo`. +1 for deprecation.

MarioBlazek (approved):

lib/Core/Site/Values/ContentInfo.php
```diff
@@ -153,11 +153,15 @@ public function __debugInfo()
...
     public function getLocations($limit = 25)
     {
+        @trigger_error('getContent() is deprecated since version 2.6 and will be removed in 3.0. Use the same method on Content object instead.', E_USER_DEPRECATED);
+
         return $this->filterLocations($limit)->getIterator();
     }
 
     public function filterLocations($maxPerPage = 25, $currentPage = 1)
     {
+        @trigger_error('getContent() is deprecated since version 2.6 and will be removed in 3.0. Use the same method on Content object instead.', E_USER_DEPRECATED);
```
emodric: Wrong method name

lib/Core/Site/Values/ContentInfo.php
```diff
@@ -153,11 +153,15 @@ public function __debugInfo()
 
     public function getLocations($limit = 25)
     {
+        @trigger_error('getContent() is deprecated since version 2.6 and will be removed in 3.0. Use the same method on Content object instead.', E_USER_DEPRECATED);
```
emodric: Wrong method name

emodric (approved):
Looks good, with minor fixes.

lib/Core/Site/Values/ContentInfo.php
pspanja: Thanks, fixed in https://github.com/netgen/ezplatform-site-api/pull/87/commits/6434f02fc210e8e8450c1f8b8589b95bd33cd321

lib/Core/Site/Values/ContentInfo.php
pspanja: Fixed in https://github.com/netgen/ezplatform-site-api/pull/87/commits/6434f02fc210e8e8450c1f8b8589b95bd33cd321

PR #90, opened 2019-03-08
Use sudo() to (lazy) load innerContent

pspanja:
If Site API Content was loaded using `sudo()`, `Content::$innerContent` should also be accessible regardless of permissions.

### DIFF
--- a/lib/Core/Site/Values/Content.php
+++ b/lib/Core/Site/Values/Content.php
@@ -364,7 +364,7 @@ final class Content extends APIContent
     {
         if ($this->innerContent === null) {
             $this->innerContent = $this->repository->sudo(
-                function() {
+                function(Repository $repository) {
                     return $this->contentService->loadContent(
                         $this->id,
                         [$this->languageCode],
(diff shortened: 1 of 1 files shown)


emodric (commented):
I'm not sure I understand. How do we know that the original content was loaded with `sudo`?

hknezevic (commented):
Tested, works as intended.

I don't think this causes any kind of security issue. 
If you cannot access the Site API Content, you cannot access the innerContent property as well.

pspanja:
@emodric 

> I'm not sure I understand. How do we know that the original content was loaded with sudo?

We don't know that, but it doesn't matter. If there is a Content, we have to assume the user has access to it, and that could have been done through `sudo()`.

emodric:
Hah, that makes sense :)

hknezevic:
@pspanja we should check if we need to add the same fix on any other inner properties (innerLocation, innerContentInfo), but we can open a separate issue for that. As far as I am concerned, this is fine

hknezevic (approved):

PR #92, opened 2019-04-01
Support for eZ Platform 2.5.0

pspanja:


lib/Core/Repository/Aggregate/Repository.php
```diff
@@ -64,9 +64,12 @@ public function setCurrentUser(UserReference $user)
         return $this->ezRepository->setCurrentUser($user);
     }
 
-    public function sudo(Closure $callback)
+    public function sudo(callable $callback, RepositoryInterface $outerRepository = null)
```
emodric, line 67: Does changing `Closure` to `callable` make this incompatible with previous versions of eZ kernel?

pspanja: No; previously the method was not in the interface and `callable` type hint will allow `Closure` instance. I updated the test setup to allow testing integration with different versions of `ezpublish-kernel`, everything works OK.

emodric: 👍

emodric (approved):

pspanja:
Thanks for the review!

PR #93, opened 2019-05-03
In case of embed view and if Location is passed in, use eZ Location

MarioBlazek:


emodric:
Can you explain why, for me and for future reference? :)

emodric:
Also  don't we have some embed views in site bundle which use site api location, which would potentially break here?

hknezevic:
@emodric seems that this happens only in case when you render the embed view for the content explicitly in the template ( render(controller('ngContent:viewAction') ), and only if you explicitly pass the locationId or location parameter to it. 

If the embed template is invoked the regular way (from the rich text ), then this will not happen, since only the content is passed on. The authorization checker for the location is invoked only if the location is passed explicitly to the embed view.

MarioBlazek:
@emodric this was an exceptional case when `embed` view was rendered and also location was passed as a parameter:

```twig
    {{ render(
        controller(
              'ng_content:viewAction', {
                    'contentId': content.id,
                    'locationId': location.id,
                    'viewType': 'embed'
               }
         )
     ) }}
```

hknezevic:
@emodric the authorization checker expects input argument to be either an instance of the eZ core Location or SPILocation

emodric:
Okay, will it mess up the outer context since this modifies a variable which is passed by reference?

pspanja:
@emodric 

> Okay, will it mess up the outer context since this modifies a variable which is passed by reference?

Actually it won't. It would happen if the object was being passed by reference, which it isn't.

Edit: explicitly passed by reference. PHP is weird :)

pspanja (changes requested):
Looks good, but I would still not overwrite method's argument. We can introduce a new variable, `$innerLocation` or `$repositoryLocation` and use it instead.

emodric:
Aren't objects passed by reference by default?

pspanja:
Kind of :) See https://3v4l.org/iMHCl

emodric:
I see. Good to know :+1:

hknezevic:
@MarioBlazek this looks better ;)

MarioBlazek:
@emodric @pspanja objects are passed as references in methods, not in functions. In functions this needs to be done explicitly.

MarioBlazek:
Thanks for pinpointing the potential issue.

emodric:
@MarioBlazek So the previous code would definitely break then?

pspanja (approved):

pspanja:
Thanks @MarioBlazek!

pspanja:
> So the previous code would definitely break then?

It would not break. Behavior is the same in both functions and methods.

emodric:
@MarioBlazek @pspanja Fight, fight! :)

pspanja:
There's really nothing to fight about, it works the way it works :)

PR #94, opened 2019-05-22
[3.0] As a developer I don't want a site to crash because of basic Content Field inconsistencies

pspanja:
Sometimes when the content model is changed or the database is inconsistent it can happen that Content Fields or ContentType FieldDefinitions are missing. In case of content model change that is a temporary situation lasting while the data is being updated in the background. It should be possible to handle this situation gracefully, instead of having it crash the whole site.

This implements graceful handling of missing Content Field or ContentType FieldDefinition by returning a `ngsurrogate` virtual type field. This field will always evaluate as empty and will be rendered through it's own Twig block, by default as an empty string. A critical-level message will be logged when this happens.

Open questions:
- ~~is this a sensible approach, should we just handle `null` in `ng_render_field` instead?~~ it's sensible
- ~~what should we render in place of such field by default?~~ empty string
- ~~do we want to crash the site in DEV environment, or should there be an option to configure that?~~ semantic config option

### To do
- [x] `Content::$fields` is exposed directly, this needs to be handled by a proxy object
- [x] ~Test missing `FieldDefinition`~ removed as not needed
- [x] Update `RelationService` to rely on the configured behavior
- [x] Write docs


emodric:
> is this a sensible approach, should we just handle null in ng_render_field instead?

Looks good to me :+1:

> what should we render in place of such field by default?

An empty string string seems most sensible, since you don't want production riddled with "random" and out of place data. If anyone needs anything else, template can be overriden to add own rendering.

> do we want to crash the site in DEV environment, or should there be an option to configure that?

Why not both? Crash by default, since the crashes in development 99% of the time indicate a deeper issue in database data which you would want to fix. Configuration can be used to temporarily disable the crashes if needed.

bundle/Resources/config/ezplatform.yml
```diff
@@ -0,0 +1,6 @@
+system:
+    default:
+        field_templates:
+            - { template: "@NetgenEzPlatformSiteApi/content_fields.html.twig", priority: 10000 }
```
emodric: Ideally, these two priorities in this file should be set to `0`, so developers overriding the field rendering do not need to look up what priority to use to correctly override the field block.

lib/Core/Site/DomainObjectMapper.php
```diff
@@ -190,6 +223,54 @@ public function mapField(APIField $apiField, SiteContent $content)
         ]);
     }
 
+    public function getNullField($identifier, SiteContent $content)
+    {
+        $message = 'nešto je trulo u državi danskoj';
```
emodric: Quite Shakespearean and fancy, but maybe put some proper message :D

lib/Resources/config/internal.yml
```diff
@@ -51,6 +51,8 @@ services:
             - '@ezpublish.api.repository'
             - '@netgen.ezplatform_site.repository.filtering_search_service'
             - '@netgen.ezplatform_site.plugins.field_type.relation_resolver.registry'
+            - '%kernel.debug%'
+            - '@logger'
```
emodric: Should be `@?logger`, since the service does not have to exist if users turn off logging.

bundle/Resources/config/ezplatform.yml
pspanja: Fixed in https://github.com/netgen/ezplatform-site-api/pull/94/commits/9ef3816b914ccd9e113cac622859da340f68b89c

lib/Core/Site/DomainObjectMapper.php
pspanja: Fixed in https://github.com/netgen/ezplatform-site-api/pull/94/commits/42632bc050a2912fb50e058f166702de74a958f3 :)

lib/Resources/config/internal.yml
pspanja: Fixed in https://github.com/netgen/ezplatform-site-api/pull/94/commits/f872cc4af9dbaab6e8e8fff10b7c2d8be6893b32

bundle/DependencyInjection/NetgenEzPlatformSiteApiExtension.php
```diff
@@ -31,6 +34,7 @@ public function load(array $configs, ContainerBuilder $container)
         $activatedBundles = array_keys($container->getParameter('kernel.bundles'));
 
         $configuration = $this->getConfiguration($configs, $container);
+        assert($configuration instanceof  Configuration);
```
emodric: Superfluous, since the method returns the `Configuration` object.

bundle/Resources/config/ezplatform.yml
```diff
@@ -0,0 +1,6 @@
+system:
+    default:
+        field_templates:
+            - { template: "@NetgenEzPlatformSiteApi/content_fields.html.twig", priority: 0 }
+        fielddefinition_settings_templates:
+            - { template: "@NetgenEzPlatformSiteApi/field_definition_settings_templates.html.twig", priority: 0 }
```
emodric: Wrong template file name, there's an extra `_templates`.

pspanja:
> Why not both? Crash by default, since the crashes in development 99% of the time indicate a deeper issue in database data which you would want to fix. Configuration can be used to temporarily disable the crashes if needed.

I added an option in semantic configuration to control this, `fail_on_missing_fields`. It defaults to `%kernel.debug%`, so we get strict behavior in DEV and tolerant behavior in PROD by default, and also a possibility to override it per siteaccess as needed.

bundle/Resources/config/ezplatform.yml
pspanja: Fixed in https://github.com/netgen/ezplatform-site-api/pull/94/commits/ce2aaafb65d50c921519c8ffb3ffda1500537d62

bundle/DependencyInjection/NetgenEzPlatformSiteApiExtension.php
pspanja: Method signature still allows for `null`, I added this to remove the warning in PHPStorm, as a replacement for inline type hint. This should be disabled in production, so it comes at no extra cost.

pspanja:
Added a to do item, `Content::$fields` is exposed directly so we'll need `ArrayAccess` or similar to replace it.

MarioBlazek (commented):
Like the idea of how this is implemented, is this stable enough for testing?

pspanja:
@MarioBlazek 
> Like the idea of how this is implemented, is this stable enough for testing?

Se my comment above, for example `{{ ng_render_field(content.fields.burek) }}` still doesn't work as the `$fields` property is exposed directly. But it can be tested through alternative access, like `{{ ng_render_field(content.getField('burek')) }}`.

bundle/DependencyInjection/NetgenEzPlatformSiteApiExtension.php
emodric: I'm more comfortable with a (inline) typehint, following the logic from PHP docs:

`Assertions should be used as a debugging feature only. You may use them for sanity-checks that test for conditions that should always be TRUE and that indicate some programming errors if not or to check for the presence of certain features like extension functions or certain system limits and features.`

You can also probably add the `@return` type hint to the method to disable the warning in PHPStorm.

This is a style preference, so you have the last say :)

pspanja: Fair point :) Changed to return type hint in https://github.com/netgen/ezplatform-site-api/pull/94/commits/a6add147ab85ac2a83e3fd75e7c1dd94bdc08e43

pspanja:
This now implements new `Fields` object, which is basically a collection of fields implementing `IteratorAggregate`, `ArrayAccess` and `Countable`. This is a change in API, as it replaces `array` type for `Content::$fields` property.

In order to keep behavior in Twig identical, I had to override Twig's `GetAttrExpression` node. The reason is because original implementation checks for property/array element existence before returning it, which results in need for `Fields::offsetExists($identifier)` to always return true. With that, checks like `{% if content.fields.kopito is defined %}` would always evaluate as true, even if the field does not exists, and would have to be replaced with `{% if content.fields.hasField('kopito') %}`. But with `GetAttrExpression` override we retain the same behavior.

Open questions:

1. `IteratorAggregate`, `ArrayAccess` and `Countable` is almost the same as `array`, but still not quite identical. Do we need v3 because of that?
2. By overriding `GetAttrExpression` we can implement this behavior for Twig access only. Do we really want it in PHP as well?

pspanja:
@MarioBlazek you can test this now :)

pspanja:
Also note: we log different messages for 2 cases, 1. when Field is missing from Content and 2. when FieldDefinition is missing from ContentType. In both cases Null field will be returned.

emodric:
I'm not sure what to think about overriding the internal parts of the Twig engine, especially the node with its `compile` method.

It will take considerable amount of time to keep an eye on the original implementation of Twig and patching it up here when something changes. Couple that with multiple future versions of Site API and multiple versions of Twig and you have a big mess on your hands.

Also, Site API supports PHP 5.6 still, and this node override uses null coalescing operator from the original implementation, so it's not compatible with PHP 5.6.

There are a couple of instances in Twig engine where the code is checking for instance of `GetAttrExpression`. I have a feeling these will fail if we just override the node.

All in all, it seems to me that this will bring more problems than potential benefits. I would rather document that if you use `if content.fields.my_field is defined` that you cannot expect this protection from crashing to work and that the only safe use of `content.fields` is to iterate over it. At worst, it will work as it worked up until now.

pspanja:
As discussed in person, it's not just about `is defined` always evaluating as true, it would also come with other undesired side effects, like from PHP `isset($content->fields['non_existent'])` always evaluating as true.

`GetAttrExpression` usage can be worked around by extending that class, but we still need to look into the history of that code if we go that direction.

pspanja:
Introduced `NullValue` in https://github.com/netgen/ezplatform-site-api/pull/94/commits/ba4a6006e1b2f9389a5d0a88724cddfda06a7ce0, a Field Value object made to be resistant to crashes if used as a Value object of another field type.

pspanja:
In https://github.com/netgen/ezplatform-site-api/pull/94/commits/1200d6eee8dafa3f5326b89a07e3706d381d6c60 added handling of relations in `RelationService`. Throwing `InvalidArgumentValue` on nonexistent field is now removed, and the logic relies on the configured behavior. With this `NgNull` relation Resolver is added, always returning an empty array of relation IDs.

pspanja:
Added documentation in https://github.com/netgen/ezplatform-site-api/pull/94/commits/3f41ffc1e27390cecee38b2595a5c69fbd764177.

pspanja:
Thanks for the review!

PR #95, opened 2019-06-02
Fix generating limitation targets

pspanja:
This continues on https://github.com/netgen/ezplatform-site-api/pull/93. The problems was really in `canRead()` method, where Site API Location was passed as a limitation target instead of a Repository Location. Since the method accepts an instance of Site API Location, previous fix was not complete.


MarioBlazek (approved):

emodric (approved):
Nice catch :+1:

PR #96, opened 2019-06-17
[2.7] Add config access to query type language expressions

pspanja:
This adds `ConfigResolver` and new `config()` method to the QueryType language expressions.
Method is a shortcut to `ConfigResolver` with default value fallback, with signature:

```
config($name, $namespace = null, $scope = null)
```

Example usage:

```yaml
queries:
    subtree:
        query_type: 'SiteAPI:Location/Children'
        max_per_page: '@=configResolver.getParameter("limit", "ngsite", "eng")'
        page: 1
        parameters:
            content_type: '@=config("content_types", "ngsite")'
```

### Todos
- [x] update docs

### DIFF
--- a/bundle/QueryType/ParameterProcessor.php
+++ b/bundle/QueryType/ParameterProcessor.php
@@ -2,6 +2,7 @@
 
 namespace Netgen\Bundle\EzPlatformSiteApiBundle\QueryType;
 
+use eZ\Publish\Core\MVC\ConfigResolverInterface;
 use Netgen\Bundle\EzPlatformSiteApiBundle\View\ContentView;
 use Symfony\Component\ExpressionLanguage\ExpressionLanguage;
 use Symfony\Component\HttpFoundation\RequestStack;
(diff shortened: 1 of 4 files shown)


MarioBlazek (approved):
Looks good.

emodric:
Any specific reason why the fallback is included, instead of just throwing when the parameter does not exist?

It could lead to subtle bugs if you mistype the parameter name, for example. Since eZ preaches that the fallback parameter should always be available (in the `default` scope), we should use that one as a fallback.

Otherwise, :+1:

pspanja:
@emodric Good point, I will remove the default value.

iherak (approved):

pspanja:
Removed in https://github.com/netgen/ezplatform-site-api/pull/96/commits/4c4466b0a8845f681d0a91d6d5780fe5f1d5b083

pspanja:
Thanks for the review!

PR #97, opened 2019-06-19
[2.7] Define type casting query param getters

pspanja:
Query/GET params, avilable through Symfony's Request object are always of string type. That can be a problem if you need to use them for QueryType conditions, which can be type hinted through `OptionResolver` component.

This adds type casting query param getter function for Query Type expression language:

- `queryParamString`
- `queryParamInt`
- `queryParamFloat`
- `queryParamBool`

Behavior is the same as with existing `queryParam` function, with the difference that value is casted to the appropriate scalar type. Usage example:

```yaml
...
    queries:
        children:
            query_type: 'SiteAPI:Location/Subtree'
            max_per_page: 10
            page: 1
            parameters:
                dapth: '@=queryParamInt("depth", 1)'
```

### Todos
- [x] tests
- [x] docs

### DIFF
--- a/bundle/QueryType/ParameterProcessor.php
+++ b/bundle/QueryType/ParameterProcessor.php
@@ -100,6 +100,66 @@ final class ParameterProcessor
         );
 
         $expressionLanguage->register(
+            'queryParamString',
+            static function () {},
+            static function ($arguments, $name, $default) {
+                /** @var \Symfony\Component\HttpFoundation\Request $request */
+                $request = $arguments['request'];
+
+                if ($request->query->has($name)) {
+                    return (string)$request->query->get($name, $default);
+                }
+
+                return $default;
+            }
+        );
+
+        $expressionLanguage->register(
+            'queryParamInt',
+            static function () {},
+            static function ($arguments, $name, $default) {
+                /** @var \Symfony\Component\HttpFoundation\Request $request */
+                $request = $arguments['request'];
+
+                if ($request->query->has($name)) {
+                    return (int)$request->query->get($name, $default);
+                }
+
+                return $default;
+            }
+        );
+
+        $expressionLanguage->register(
+            'queryParamFloat',
+            static function () {},
+            static function ($arguments, $name, $default) {
+                /** @var \Symfony\Component\HttpFoundation\Request $request */
+                $request = $arguments['request'];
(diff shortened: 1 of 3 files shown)


MarioBlazek (approved):
This is a welcoming addition. Like it.

bundle/QueryType/ParameterProcessor.php
```diff
@@ -99,6 +99,66 @@ static function ($arguments, $name, $default) {
...
+
+        $expressionLanguage->register(
+            'queryParamBool',
+            static function () {},
+            static function ($arguments, $name, $default) {
+                /** @var \Symfony\Component\HttpFoundation\Request $request */
+                $request = $arguments['request'];
+
+                if ($request->query->has($name)) {
+                    return (bool)$request->query->get($name, $default);
```
emodric: You can use `$request->query->getBoolean()`

bundle/QueryType/ParameterProcessor.php
```diff
@@ -99,6 +99,66 @@ static function ($arguments, $name, $default) {
...
+
+        $expressionLanguage->register(
+            'queryParamInt',
+            static function () {},
+            static function ($arguments, $name, $default) {
+                /** @var \Symfony\Component\HttpFoundation\Request $request */
+                $request = $arguments['request'];
+
+                if ($request->query->has($name)) {
+                    return (int)$request->query->get($name, $default);
```
emodric: You can use `$request->query->getInt()`

emodric (approved):
Apart from minor nitpicks :+1:

bundle/QueryType/ParameterProcessor.php
pspanja: Fixed in https://github.com/netgen/ezplatform-site-api/pull/97/commits/f72f5ecae52253ce294122b73d769c9a0157745a

bundle/QueryType/ParameterProcessor.php
pspanja: Fixed in https://github.com/netgen/ezplatform-site-api/pull/97/commits/f72f5ecae52253ce294122b73d769c9a0157745a

pspanja:
Thanks for the review!

PR #109, opened 2019-07-08
Add default view matcher, configuring the view with fallback templates

emodric:
It turns out there is a view matcher in eZ that provides some default templates for rarely used view types like `embed-inline` and so on.

This configures the same view matcher for Site API specific view (`Netgen\Bundle\EzPlatformSiteApiBundle\View\ContentView`).

Otherwise, you would get an exception: `No view template was set to render the view with the 'embed-inline' view type. Check your view configuration.`

Tested on eZ kernel v8.

### DIFF
--- a/bundle/Resources/config/view.yml
+++ b/bundle/Resources/config/view.yml
@@ -43,3 +43,18 @@ services:
             - [setContainer, ["@service_container"]]
             - [setMatchConfig, [$content_view$]]
         public: false
+
+    netgen.ezplatform_site.ngcontent_view_provider.default_configured:
+        class: eZ\Bundle\EzPublishCoreBundle\View\Provider\Configured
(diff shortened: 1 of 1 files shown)


pspanja:
I think here we should use [DynamicallyConfiguredMatcherFactoryDecorator](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/MVC/Symfony/Matcher/DynamicallyConfiguredMatcherFactoryDecorator.php) to avoid dynamic settings.

emodric:
:+1: Will update!

emodric:
Unfortunatelly, that class is not available in eZ kernel 7, so it will have to wait untill we make a release specific to eZ kernel 8.

pspanja:
Makes sense, kernel 8 will break things anyway 👍

PR #114, opened 2019-07-17
Declare that get Location methods on objects return arrays

pspanja:
Methods `Content::getLocations()`, `Location::getChildren()` and `Location::getSiblings()` were hinting to return an array, but were actually returning an instance if `ArrayIterator`. This updates the implementation to return an array and adds the `array` return type declaration.

Note that this is technically a BC break.

### To do
- [x] add test coverage


emodric:
Looks good :+1:

emodric (approved):

pspanja:
Thanks!

PR #115, opened 2019-07-19
Refactor base Controller

pspanja:
This removes extension of `eZ\Bundle\EzPublishCoreBundle\Controller` from base Site API controller. Instead, the controller now extends `Symfony\Bundle\FrameworkBundle\Controller\Controller` directly. This is done because of mistake where `getRootLocation()` is reused to return Site API Location object, while the base method returns Repository Location.

All 3 service getter methods that were previously inherited from `eZ\Bundle\EzPublishCoreBundle\Controller` are now implemented here.

This is a BC break because any type hints for `eZ\Bundle\EzPublishCoreBundle\Controller` where Site API base controller is used will fail. However, it's highly unlikely anyone uses that.

### DIFF
--- a/bundle/Controller/Controller.php
+++ b/bundle/Controller/Controller.php
@@ -4,12 +4,15 @@ declare(strict_types=1);
 
 namespace Netgen\Bundle\EzPlatformSiteApiBundle\Controller;
 
-use eZ\Bundle\EzPublishCoreBundle\Controller as BaseController;
+use eZ\Publish\API\Repository\Repository;
+use eZ\Publish\Core\MVC\ConfigResolverInterface;
+use eZ\Publish\Core\MVC\Symfony\Templating\GlobalHelper;
 use eZ\Publish\Core\QueryType\QueryTypeRegistry;
 use Netgen\EzPlatformSiteApi\API\Site;
 use Netgen\EzPlatformSiteApi\API\Values\Location;
 use Netgen\EzPlatformSiteApi\Core\Traits\PagerfantaTrait;
 use Netgen\EzPlatformSiteApi\Core\Traits\SearchResultExtractorTrait;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller as BaseController;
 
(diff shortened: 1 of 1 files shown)


bundle/Controller/Controller.php
```diff
@@ -4,12 +4,15 @@
...
-use eZ\Bundle\EzPublishCoreBundle\Controller as BaseController;
+use eZ\Publish\API\Repository\Repository;
+use eZ\Publish\Core\MVC\ConfigResolverInterface;
+use eZ\Publish\Core\MVC\Symfony\Templating\GlobalHelper;
 use eZ\Publish\Core\QueryType\QueryTypeRegistry;
 use Netgen\EzPlatformSiteApi\API\Site;
 use Netgen\EzPlatformSiteApi\API\Values\Location;
 use Netgen\EzPlatformSiteApi\Core\Traits\PagerfantaTrait;
 use Netgen\EzPlatformSiteApi\Core\Traits\SearchResultExtractorTrait;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller as BaseController;
```
emodric, line 15: This should be `Symfony\Bundle\FrameworkBundle\Controller\AbstractController` as the one here is deprecated in 4.2.

https://symfony.com/blog/new-in-symfony-4-2-important-deprecations#deprecated-the-base-controller-class

`AbstractController` is available in Symfony 3.4, so no issues there.

emodric: Hm... `AbstractController` does restrict the usage of `$this->get()`, but there is a way around it too, as far as I can see.

pspanja: Updated in https://github.com/netgen/ezplatform-site-api/pull/115/commits/af471d846009e3e255488d1b015e936d69f2e285

pspanja: Missed your 2nd comment, not reverted the commit from above as discussed in person. We'll update for 4.2 later.

PR #117, opened 2019-07-21
[3.0] As a developer I want to render Content view without using a subrequest

pspanja:
This implements experimental support for rendering Content views without executing a subrequest.

Performance impact of subrequests in production mode is not significant, but that is also not where developers spend most of their time. Because of the profiling, performance impact in debug mode can be huge. In one test we executed, a page with 40 subrequests that took ~3 seconds to render, needed only half as much when subrequests were replaced.

For most use cases it is not really necessary to render a view through a subrequest, and providing an alternative, more performant way is a chance to improve developer's experience.

**Note that this is not intended to be a complete replacement for subrequests, because it does not dispatch MVC events it can be safely used only for those views that do not depend on them. Also, it should be used only with configured views (`ngcontent_view` semantic configuration):**

```yaml
ezpublish:
    system:
        frontend_group:
            ngcontent_view:
                line:
                    article:
                        template: "@ezdesign/content/line/article.html.twig"
                        controller: custom_controller:viewAction
                        match:
                            Identifier\ContentType: article
```

Twig function `ng_view_content` will accept both Content and Location objects:

```twig
{{ ng_view_content(articleContent, 'line', {'name': 'value'} }}
{{ ng_view_content(articleLocation, 'line', {'name': 'value'} }}
```

Rendering a nonexistent view will intentionally result in a fatal error. That is different from how executing a subrequest behaves, where it will silently produce an empty string.

Custom controllers are also supported and that is probably the part that needs most testing.

### Todos
- [x] write some documentation

### DIFF
--- a/bundle/Templating/Twig/Extension/ContentViewRuntime.php
+++ b/bundle/Templating/Twig/Extension/ContentViewRuntime.php
@@ -1,0 +1,212 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Netgen\Bundle\EzPlatformSiteApiBundle\Templating\Twig\Extension;
+
+use function call_user_func_array;
+use eZ\Publish\API\Repository\Values\ValueObject;
+use eZ\Publish\Core\MVC\Symfony\View\Renderer;
+use LogicException;
+use Netgen\Bundle\EzPlatformSiteApiBundle\View\Builder\ContentViewBuilder;
+use Netgen\Bundle\EzPlatformSiteApiBundle\View\ContentView;
+use Netgen\EzPlatformSiteApi\API\Values\Content;
+use Netgen\EzPlatformSiteApi\API\Values\Location;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Controller\ArgumentResolverInterface;
+use Symfony\Component\HttpKernel\Controller\ControllerReference;
+use Symfony\Component\HttpKernel\Controller\ControllerResolverInterface;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+
+/**
+ * Twig extension runtime for content rendering (view).
+ */
+class ContentViewRuntime
+{
+    /**
+     * @var \Symfony\Component\HttpFoundation\RequestStack
+     */
+    private $requestStack;
+
+    /**
+     * @var \Symfony\Component\HttpKernel\Controller\ControllerResolverInterface
+     */
+    private $controllerResolver;
+
+    /**
+     * @var \Symfony\Component\HttpKernel\Controller\ArgumentResolverInterface
+     */
+    private $argumentResolver;
+
+    /**
+     * @var \Netgen\Bundle\EzPlatformSiteApiBundle\View\Builder\ContentViewBuilder
+     */
+    private $viewBuilder;
+
+    /**
+     * @var \eZ\Publish\Core\MVC\Symfony\View\Renderer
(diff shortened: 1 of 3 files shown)


bundle/Templating/Twig/Extension/ContentViewRuntime.php
```diff
@@ -0,0 +1,212 @@
...
+        string $viewType,
+        ValueObject $contentOrLocation,
+        array $parameters = []
+    ): string {
+        $this->logger->critical('Twig function "ng_view_content" is experimental. Please report any issues on https://github.com/netgen/ezplatform-site-api/issues');
+
+        $content = $this->getContent($contentOrLocation);
+
+        $view = $this->viewBuilder->buildView([
+            'viewType' => $viewType,
```
emodric, line 97: This might not be a good idea. These will end up on Sentry in production mode. A comment in phpdoc might be enough.

bundle/Templating/Twig/Extension/ContentViewRuntime.php
```diff
@@ -0,0 +1,212 @@
...
+     * @param array $parameters
+     *
+     * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentException
+     * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentType
+     * @throws \eZ\Publish\Core\Base\Exceptions\UnauthorizedException
+     *
+     * @return string The HTML markup
+     */
+    public function renderContentView(
+        string $viewType,
```
emodric: How about placing content/location first? We are rendering a content in a specific view, not the other way around.

emodric (changes requested):
Left a few remarks, otherwise looks good ( as far as I can judge from the beach :) )

bundle/Templating/Twig/Extension/ContentViewRuntime.php
emodric: Went on the wrong line, this was supposed to comment a critical logger call.

pspanja: Hm, do you mean about throwing `LogicException` below?

bundle/Templating/Twig/Extension/ContentViewRuntime.php
pspanja: Yes indeed, will change 👍

bundle/Templating/Twig/Extension/ContentViewRuntime.php
pspanja: OK, updated as suggested in https://github.com/netgen/ezplatform-site-api/pull/117/commits/4b98a242dcfb3a2b92f967514c44ae7edf473155

bundle/Templating/Twig/Extension/ContentViewRuntime.php
pspanja: Updated in https://github.com/netgen/ezplatform-site-api/pull/117/commits/6bb212e8833c915a0a0085a238d0c79eab6be04d

pspanja:
Thanks, enjoy the sea 😄

emodric (approved):

hknezevic (approved):
Nice! 
I am eager to test this on some heavy duty sites in development. Landing pages with 50+ items in various blocks sometimes take up to 10 seconds for a reload in development. This should significantly improve the experience for the developers.

Other than the stuff Edi already commented and you fixed, I don't see any issue in the code. In combination with the existing Site API features (related content fetches, location/content being available out of the box in lazy loaded properties), we will no longer be dependent on rendering view actions for content IDs and location IDs in most of the cases.

pspanja:
> we will no longer be dependent on rendering view actions for content IDs and location IDs in most of the cases.

Great, that was one doubt I had about it, but we shouldn't need it with all the stuff already in place 👍 

I'm going to merge this now so we can test it in action.

PR #123, opened 2019-07-23
Add support for eZ Platform v3 version of RepositoryFactory::buildRepository

emodric:
https://github.com/ezsystems/ezpublish-kernel/pull/2688 updated the signature of `buildRepository` method to add a new argument.

The service is added with `@?` annotation in order not to crash on eZ Platform 2.5 where the service does not exist.


pspanja:
Thanks @emodric

PR #124, opened 2019-07-24
Add configuration for surrogate field rendering

pspanja:
This adds siteaccess aware configuration option `render_missing_field_info`, which controls whether surrogate field will be rendered as an empty string or it will render useful debug information.

### DIFF
--- a/bundle/Resources/config/services.yml
+++ b/bundle/Resources/config/services.yml
@@ -11,7 +11,7 @@ parameters:
     netgen_ez_platform_site_api.default.override_url_alias_view_action: false
     netgen_ez_platform_site_api.default.use_always_available_fallback: true
     netgen_ez_platform_site_api.default.fail_on_missing_fields: '%kernel.debug%'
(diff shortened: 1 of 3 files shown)


bundle/Resources/config/services.yml
```diff
@@ -11,6 +11,7 @@ parameters:
     netgen_ez_platform_site_api.default.override_url_alias_view_action: false
     netgen_ez_platform_site_api.default.use_always_available_fallback: true
     netgen_ez_platform_site_api.default.fail_on_missing_fields: '%kernel.debug%'
+    netgen_ez_platform_site_api.default.render_missing_field_info: true
```
emodric: I think that this should be disabled by default. In debug mode we have an exception by default, and in production mode we don't want this info displayed in any case.

pspanja: Makes sense, updated in https://github.com/netgen/ezplatform-site-api/pull/124/commits/287524d7cded67dc03e3dc8e43a59796cc365c9c

emodric (approved):

PR #125, opened 2019-07-31
Add routers that generate links for Site API content and location

emodric:
No tests for now, not sure how to test this properly.

If using mocks, creating value objects for passing to mocks is complicated due to various dependencies.


pspanja:
It's OK if only manually tested, we can improve later. But please move it to `bundle/Routing/`, we already have a router there :)

emodric:
Done!

pspanja:
Great, thanks!

PR #129, opened 2019-08-12
Implement ParamConverters for SiteAPI value objects

MarioBlazek:
This PR implements ParamConverters for SiteAPI `Content` and `Location` value objects.


bundle/Converter/SiteParamConverter.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        if (!$request->attributes->has($this->getPropertyName())) {
+            return false;
+        }
+        $valueObjectId = $request->attributes->get($this->getPropertyName());
+        if (!$valueObjectId && $configuration->isOptional()) {
+            return false;
+        }
+
+        $request->attributes->set($configuration->getName(), $this->loadValueObject($valueObjectId));
```
emodric, line 48: ID should probaby be cast to `int`, or use `$request->attributes->getInt()` to get the proper type.

bundle/DependencyInjection/Configuration.php
```diff
@@ -50,5 +50,10 @@ protected function addConfiguration(ArrayNodeDefinition $rootNode): void
             ->booleanNode('render_missing_field_info')
                 ->info('Whether to render useful debug information in place of a missing field')
             ->end();
+        $systemNode
+            ->booleanNode('use_param_converters')
```
emodric: Why do we need this? Isn't the purpose of `supports` method to declare what typehints param converter supports?

Moreover, I don't think this "overrides" eZ param converters. They can live side by side without issues.

emodric (changes requested):

bundle/Converter/SiteParamConverter.php
MarioBlazek: @emodric comment addressed.

emodric (approved):

emodric:
Looks good!

We should add `"sensio/framework-extra-bundle": "^5.2"` bundle to `composer.json` though.

bundle/Converter/LocationParamConverter.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Netgen\Bundle\EzPlatformSiteApiBundle\Converter;
+
+use Netgen\EzPlatformSiteApi\API\Values\Location;
+
+class LocationParamConverter extends SiteParamConverter
```
emodric: Missing `final`.

pspanja (approved):

pspanja:
Thanks @MarioBlazek, this was long overdue 👍

bundle/Resources/config/services.yml
```diff
@@ -80,3 +80,17 @@ services:
             - "@?router.request_context"
         tags:
             - { name: router, priority: 200 }
+
+    netgen.ezplatform_site.param_converter.location:
+        class: Netgen\Bundle\EzPlatformSiteApiBundle\Converter\ContentParamConverter
+        arguments:
+            - '@netgen.ezplatform_site.site'
```
emodric, line 87: @MarioBlazek This should've been the `@netgen.ezplatform_site.load_service` instead of site.

I'm getting: `Type error: Argument 1 passed to Netgen\Bundle\EzPlatformSiteApiBundle\Converter\SiteParamConverter::__construct() must implement interface Netgen\EzPlatformSiteApi\API\LoadService, instance of Site_221043a given`

bundle/Resources/config/services.yml
```diff
@@ -80,3 +80,17 @@ services:
...
+        tags:
+            - { name: request.param_converter, priority: -1 }
+
+    netgen.ezplatform_site.param_converter.content:
+        class: Netgen\Bundle\EzPlatformSiteApiBundle\Converter\LocationParamConverter
+        arguments:
+            - '@netgen.ezplatform_site.site'
```
emodric, line 94: Here too, should be `@netgen.ezplatform_site.load_service`.

PR #130, opened 2019-08-15
Default view templates

pspanja:
This adds Site API versions of default view templates [from eZ Platform kernel](https://github.com/ezsystems/ezpublish-kernel/tree/master/eZ/Bundle/EzPublishCoreBundle/Resources/views/default/content). In eZ Platform some of these (asset image, embed image, text linked) are only used from content field templates. Since we reuse the default content field templates, it was not strictly necessary to add all of them, I did it for the sake of completeness in case someone does depend on them being available.

All of these are now configured with the default template view provider, which previously used the templates/configuration from eZ Platform.

Additionally I improved the default full view template to be more useful.

### DIFF
--- a/bundle/Resources/config/default_settings.yml
+++ b/bundle/Resources/config/default_settings.yml
@@ -6,13 +6,13 @@ parameters:
     netgen_ez_platform_site_api.default.render_missing_field_info: false
 
     # Default view templates
-    netgen.ezplatform_site.default_view_templates.content.asset_image: 'NetgenEzPlatformSiteApiBundle:default:content/asset_image.html.twig'
-    netgen.ezplatform_site.default_view_templates.content.embed: 'NetgenEzPlatformSiteApiBundle:default:content/embed.html.twig'
-    netgen.ezplatform_site.default_view_templates.content.embed_image: 'NetgenEzPlatformSiteApiBundle:default:content/embed_image.html.twig'
-    netgen.ezplatform_site.default_view_templates.content.embed_inline: 'NetgenEzPlatformSiteApiBundle:default:content/embed_inline.html.twig'
-    netgen.ezplatform_site.default_view_templates.content.full: 'NetgenEzPlatformSiteApiBundle:default:content/full.html.twig'
-    netgen.ezplatform_site.default_view_templates.content.line: 'NetgenEzPlatformSiteApiBundle:default:content/line.html.twig'
-    netgen.ezplatform_site.default_view_templates.content.text_linked: 'NetgenEzPlatformSiteApiBundle:default:content/text_linked.html.twig'
+    netgen.ezplatform_site.default_view_templates.content.asset_image: "@@NetgenEzPlatformSiteApi/default/content/asset_image.html.twig"
+    netgen.ezplatform_site.default_view_templates.content.embed: "@@NetgenEzPlatformSiteApi/default/content/embed.html.twig"
+    netgen.ezplatform_site.default_view_templates.content.embed_image: "@@NetgenEzPlatformSiteApi/default/content/embed_image.html.twig"
+    netgen.ezplatform_site.default_view_templates.content.embed_inline: "@@NetgenEzPlatformSiteApi/default/content/embed_inline.html.twig"
+    netgen.ezplatform_site.default_view_templates.content.full: "@@NetgenEzPlatformSiteApi/default/content/full.html.twig"
+    netgen.ezplatform_site.default_view_templates.content.line: "@@NetgenEzPlatformSiteApi/default/content/line.html.twig"
+    netgen.ezplatform_site.default_view_templates.content.text_linked: "@@NetgenEzPlatformSiteApi/default/content/text_linked.html.twig"
     # List of content type identifiers to display as image when embedded
     netgen.ezplatform_site.content_view.image_embed_content_type_identifiers: ['image']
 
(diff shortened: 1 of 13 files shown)


bundle/Resources/views/default/content/embed_image.html.twig
```diff
@@ -0,0 +1,13 @@
+{% set image_field_identifier = ez_content_field_identifier_first_filled_image(content) %}
```
emodric: This function name is new in eZ Platform v3. In v2 it was called `ez_first_filled_image_field_identifier`.

I have not found a sane way to implement compatibility between them :/

bundle/Resources/views/default/content/asset_image.html.twig
```diff
@@ -0,0 +1,5 @@
+{% set image_field_identifier = ez_content_field_identifier_image_asset() %}
```
emodric: Same as previous comment, this name only exists in v3. In v2 it was called `ez_image_asset_content_field_identifier`.

Ref: https://doc.ezplatform.com/en/master/releases/ez_platform_v3.0_deprecations/#functions-renamed

bundle/Resources/config/view.yml
```diff
@@ -45,5 +42,5 @@ services:
             - 'eZ\Publish\Core\MVC\Symfony\Matcher\ContentBased'
         calls:
             - [setContainer, ["@service_container"]]
-            - [setMatchConfig, [$content_view_defaults$]]
+            - [setMatchConfig, [$ng_content_view_defaults$]]
```
emodric: Maybe call this `ngcontent_view_defaults` to keep it consistent with `ngcontent_view` name for match rules?

bundle/Resources/config/default_settings.yml
```diff
@@ -0,0 +1,52 @@
+parameters:
+    # Defaults for Site API bundle semantic configuration
+    netgen_ez_platform_site_api.default.override_url_alias_view_action: false
+    netgen_ez_platform_site_api.default.use_always_available_fallback: true
+    netgen_ez_platform_site_api.default.fail_on_missing_fields: '%kernel.debug%'
+    netgen_ez_platform_site_api.default.render_missing_field_info: false
+
+    # Default view templates
+    netgen.ezplatform_site.default_view_templates.content.asset_image: 'NetgenEzPlatformSiteApiBundle:default:content/asset_image.html.twig'
```
emodric: You should use Twig notation to reference the templates, bundle notation is deprecated. In this instance it would be `@@NetgenEzPlatformSiteApi/default/content/asset_image.html.twig`

(double `@` is intentional to prevent the template name to be parsed as a service)

bundle/Resources/views/default/content/full.html.twig
```diff
@@ -1,12 +1,86 @@
...
+                    <dd>
+                        {% if field.description is not empty %}
+                            {{ field.description|default('<code>[empty]</code>') }}
+                        {% else %}
+                            <code class="empty">[empty]</code>
+                        {% endif %}
+                    </dd>
+                    <dt>Contents</dt>
+                    <dd>
+                        {% if not field.isEmpty() %}
```
emodric: Maybe write this `{% if not field.empty %}` to make it a bit leaner?

bundle/Resources/views/default/content/embed_image.html.twig
pspanja: Maybe `ng_content_field_identifier_first_filled_image`? 😂

emodric: :D

Actually, now that you mention it, I remembered I did implement the exact same BC layer in Layouts & eZ Platform integration: https://github.com/netgen-layouts/layouts-ezplatform/commit/7d21b3655f7b2308535b644f3bf48f4029ffbe20

pspanja: I think I'll go the same way. One alternative would be to implement `function_exists`, but it does not seem the better approach.

pspanja: It turns out this one can be easily written in Twig, updated in https://github.com/netgen/ezplatform-site-api/pull/130/commits/b1f4bfb60b47a08b074680f8bc1ba76893c88fac

bundle/Resources/views/default/content/asset_image.html.twig
pspanja: Rewrote to Twig in https://github.com/netgen/ezplatform-site-api/pull/130/commits/119c3fffde0c4099ac2740cda386f8f5bcd2adc4 ¯\\\_(ツ)_/¯

bundle/Resources/config/view.yml
pspanja: Updated in https://github.com/netgen/ezplatform-site-api/pull/130/commits/c6c095ea93bd4741180f08d7e925bb1f939e9f57, I guess we are now stuck with it.

bundle/Resources/config/default_settings.yml
pspanja: Fixed in https://github.com/netgen/ezplatform-site-api/pull/130/commits/1423efe0bd75f76b261a34a38f7ad846a295ed38

bundle/Resources/views/default/content/full.html.twig
pspanja: Updated in https://github.com/netgen/ezplatform-site-api/pull/130/commits/4823ad7ee41d4a3499f5bdb5354956e0cb052bfb

bundle/Resources/config/view.yml
```diff
@@ -45,7 +42,8 @@ services:
             - '@ezpublish.api.repository'
             - 'eZ\Publish\Core\MVC\Symfony\Matcher\ContentBased'
             - '@ezpublish.config.resolver'
-            - 'content_view_defaults'
+            - 'ngcontent_view_defaults'
         calls:
             - [setContainer, ["@service_container"]]
+
```
emodric: The blank line here looks like a mistake?

emodric (approved):
Apart from a single nitpick :+1:

bundle/Resources/config/view.yml
pspanja: Indeed, fixed in https://github.com/netgen/ezplatform-site-api/pull/130/commits/d8692fb1d11e2f3c2e18644bd0f3f9e5cbb27784

pspanja:
Thanks @emodric!

PR #131, opened 2019-08-15
Fix draft preview

pspanja:
This fixes the draft preview, both in Netgen Admin UI and eZ Platform Admin UI. The preview was not working because the view Matcher config for the preview siteaccess was not set to the Matcher. This is now done through our local implementation of the Matcher factory, which uses ConfigResolver directly, as was recently introduced in kernel in https://github.com/ezsystems/ezpublish-kernel/pull/2673 (for eZ Platform v3 only).

This also gets us rid of using dynamic settings.


emodric (approved):

pspanja:
Thanks @emodric!

PR #133, opened 2019-09-25
Allow eZ value objects when rendering content without subrequests

emodric:


bundle/Templating/Twig/Extension/ContentViewRuntime.php
```diff
@@ -127,29 +139,40 @@ private function viewMatched(ContentView $contentView): bool
...
 
         if ($contentOrLocation instanceof Content) {
             return $contentOrLocation->mainLocation;
         }
 
+        if ($contentOrLocation instanceof APIContent) {
+            // View builder will not load the main location if it is not provided,
+            // this makes sure it is available to the template
+            try {
+                return $this->locationService->loadLocation($contentOrLocation->contentInfo->mainLocationId);
```
pspanja: This will be a problem if main Location does not exist and the ID parameter is hinted as integer.

emodric: Right. Fixed.

I've removed the `try...catch` now since non-existent location in case content references it via main location ID is now a separate issue, pointing to data inconsistencies.

pspanja (approved):

pspanja:
Thanks @emodric!

PR #141, opened 2019-10-18
Content render events

emodric:
Since direct content rendering does not execute event listeners for subrequests ( since there are no subrequests :) ), some important stuff might be missing in the final page rendering.

For this reason, this adds an event that can be used to bring back custom functionality when rendering content without subrequests.

Additionally, this also adds a listener for said event which tags the master request with `xkey` tags relevant to content rendered without subrequests, which is normally done via `EzSystems\PlatformHttpCacheBundle\EventSubscriber\HttpCacheResponseSubscriber`. Not having the tags, and not having a working HTTP cache clear was the whole motivation behind this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ branches:
     only:
         - master
         - 2.7-release
+        - 3.0-release
 
 # make sure to update composer to latest available version
 before_install:
--- a/bundle/EventListener/ContentTaggerListener.php
+++ b/bundle/EventListener/ContentTaggerListener.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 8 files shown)


pspanja:
Great catch! I rebased this to `3.0-release`, since it's a bug fix.

pspanja:
Thanks @emodric

.travis.yml
```diff
@@ -30,6 +30,7 @@ branches:
     only:
         - master
         - 2.7-release
+        - 3.0-release
```
emodric, line 33: You can use a regex here :) e.g. `- /^\d.\d+-release$/`

pspanja: Changed in 3ceac7eaf692bfea54ab11786a172930a3b03203 👍

PR #142, opened 2019-10-22
Reset ezsettings parameters only if they do not exist

emodric:
Consider the following:

```
ezpublish:
    system:
        default:
            ngcontent_view:
                full:
                    article:
                        template: "@ezdesign/content/full/article.html.twig"
                        match:
                            Identifier\ContentType: article
```

Since eZ kernel bundle is most of the time activated before Site API bundle, this semantic configuration is processed and merged, and the internal eZ kernel parameter for the above config (`ezsettings.default.ngcontent_view`) is set, all before Site API DI extension is ran.

When Site API DI extension is ran, this causes the extension to overwrite the internal eZ kernel parameter to an empty value, even if eZ kernel already processed and merged the config.

This PR fixes this so the default value of the internal parameter is set only if there no parameter in the container yet. The basic issue here is that the semantic config is part of the eZ kernel bundles (or rather, DI extension with `ezpublish` alias) and not part of the Site API DI extension, making this a chicken or the egg situation.

Same principle applies to `ng_named_query` config.

### DIFF
--- a/bundle/DependencyInjection/NetgenEzPlatformSiteApiExtension.php
+++ b/bundle/DependencyInjection/NetgenEzPlatformSiteApiExtension.php
@@ -57,6 +57,14 @@ class NetgenEzPlatformSiteApiExtension extends Extension implements PrependExten
         $loader->load('services.yml');
         $loader->load('view.yml');
 
+        if (!$container->hasParameter('ezsettings.default.ngcontent_view')) {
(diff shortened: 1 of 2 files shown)


pspanja:
Great catch again, I also rebased this to `3.0-release`.

pspanja (approved):

PR #143, opened 2019-10-27
Cleanup setting default config values

pspanja:
This removes setting default values introduced in https://github.com/netgen/ezplatform-site-api/commit/12ca1bae062e27258a438b74e7cb68d6708ea2e1, since it seems it was about the same problem fixed in https://github.com/netgen/ezplatform-site-api/pull/142.

Additionally it extracts setting default values into a separate documented method.


emodric:
Nice catch :)

emodric (approved):

pspanja:
Thanks!

PR #144, opened 2019-10-27
Remove configuration for the default full view

pspanja:
This removes default full view configuration because some Content does not have a full view.
For some cases we depend on the default behavior, which redirects such route to the homepage.


emodric (approved):

pspanja:
Thanks @emodric!

PR #145, opened 2019-10-27
[3.1] Implement Named Objects

pspanja:
This implements Named Objects feature, which is a way to configure Content, Location and Tag objects by name to ID, in a siteaccess-aware way, and a way to access them from PHP, Twig and view configuration.

Configuration example:
```yaml
netgen_ez_platform_site_api:
  system:
    frontend_group:
      fail_on_missing_fields: false
      render_missing_field_info: true
      named_object:
        content:
          certificate: 42
          site_info:
            remote_id: 'af43sad'
        location:
          home: 2
        tag:
          tag_container:
            id: 43
          topics: 'ec32571'
```

The example shows that it's possible to configure the object either by ID or by remote ID. A shortcut syntax is provided, where ID is expected to be of integer and remote ID of string type.

Usage from PHP:
```php
use Netgen\Bundle\EzPlatformSiteApiBundle\NamedObject\Provider;

class Service
{
    /**
     * @var \Netgen\Bundle\EzPlatformSiteApiBundle\NamedObject\Provider
     */
    private $namedObjectProvider;

    public function __construct(Provider $namedObjectProvider)
    {
        $this->namedObjectProvider = $namedObjectProvider;
    }

    public function work()
    {
        $homepage = $this->namedObjectProvider->getLocation('home');

        ...
    }
}
```

Usage from Twig:
```twig
{% set homepage = ng_named_location('home') %}

{% if homepage is not null %}
    <h1>{{ homepage.content.name }}</h1>
{% endif %}
```

Usage from view configuration:
```yaml
ezpublish:
  system:
    frontend_group:
      ngcontent_view:
        full:
          ng_blog_post:
            template: "@ezdesign/content/full/ng_blog_post.html.twig"
            queries:
              top_locations:
                query_type: SiteAPI:Location/Children
                max_per_page: 10
                parameters:
                  location: '@=namedLocation("home")'
            match:
              Identifier\ContentType: ng_blog_post
```
### To dos
- [x] ~~given https://wiki.php.net/rfc/union_types_v2, check if it's safe to assume that ID will always be `integer`~~ It's safe
- [x] validate object name as Twig variable name
- [x] write documentation


pspanja:
Open questions:
1. There is a PHP API to check if the named object exists (eg. `Provider::hasContent(string $name)` and others). This is not replicated in Twig, instead for `ng_named_content` and others `null` is returned if the named object is not configured. This is done because in Twig checking if something is defined is the usual approach. Is that the correct assumption / do we need `ng_has_named_content` in Twig?

Edit: concluded to keep it like it is, as that comes natural in Twig.

emodric:
Hm... It seems that `null` evaluates as defined: https://twigfiddle.com/5j31f2 , so maybe hasser makes sense, if nothing, for consistency between APIs. But then again, retreiving the named content/location/tag and then checking for `null` also works, and it's more readable and concise.

emodric:
Also, `named_object` config looks kinda weird, considering the object is the old name for content in eZ legacy, so you have `named_object.content` or `named_object.location` config.

Don't have an alternate name yet, though :)

pspanja:
> But then again, retreiving the named content/location/tag and then checking for `null` also works, and it's more readable and concise.

Right, that still seems to come more natural for Twig.

MarioBlazek:
+1 for hasser method.

bundle/DependencyInjection/Configuration.php
```diff
@@ -57,5 +57,108 @@ protected function addConfiguration(ArrayNodeDefinition $rootNode): void
...
+                        ->end()
+                    ->end()
+                    ->arrayNode('tag')
+                        ->info('Tags by name')
+                        ->useAttributeAsKey('name')
+                        ->normalizeKeys(false)
+                        ->arrayPrototype()
+                            ->info('Tag ID or remote ID')
+                            ->beforeNormalization()
+                                ->ifTrue(static function ($v) {return !\is_array($v);})
```
MarioBlazek, line 147: Nice one :)

hknezevic:
> Hm... It seems that `null` evaluates as defined: https://twigfiddle.com/5j31f2 , so maybe hasser makes sense, if nothing, for consistency between APIs. But then again, retreiving the named content/location/tag and then checking for `null` also works, and it's more readable and concise.

Null most certainly evaluates as defined, only the existence of a named variable is evaluated. That is why we always pair 'is defined' and 'is not empty' in the checks.

From my perspective, this provides good enough dev experience. I don't mind doing a fetch in twig with the ng_named_content() and then checking if it is not empty. It's what we regularly do with any variable passed to Twig anyway.

hknezevic:
@pspanja +1 for the PR, beautiful and very useful feature. 
I don't see any objections with the code, and I'm eager to test it on the projects ASAP.

hknezevic (approved):

bundle/DependencyInjection/Configuration.php
```diff
@@ -57,5 +57,115 @@ protected function addConfiguration(ArrayNodeDefinition $rootNode): void
...
+            if (\is_int($v)) {
+                return ['id' => $v];
+            }
+
+            return ['remote_id' => $v];
+        };
+
+        /* @noinspection NullPointerExceptionInspection */
+        $systemNode
+            ->arrayNode('named_object')
```
emodric: If we're gonna keep this one, maybe a better name would be in plural: `named_objects`?

pspanja: Agreed, updated in https://github.com/netgen/ezplatform-site-api/pull/145/commits/b1ef7a9a4f32a3ff4221e2dbe6ad67e3a5a7ec32

bundle/NamedObject/Provider/Loading.php
```diff
@@ -0,0 +1,203 @@
...
+    /**
+     * {@inheritdoc}
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException
+     */
+    public function getTag(string $name): Tag
+    {
+        if ($this->tagsService === null) {
+            throw new RuntimeException('Missing Netgen TagsBundle package (netgen/TagsBundle)');
```
emodric: nitpick: lowercase `netgen/tagsbundle` :) Especially since non-lowercase package names are deprecated by Composer.

bundle/Resources/config/default_settings.yml
```diff
@@ -4,6 +4,7 @@ parameters:
     netgen_ez_platform_site_api.default.use_always_available_fallback: true
     netgen_ez_platform_site_api.default.fail_on_missing_fields: '%kernel.debug%'
     netgen_ez_platform_site_api.default.render_missing_field_info: false
+    netgen_ez_platform_site_api.default.named_object: []
```
emodric: Maybe also call it with plural name to keep it in sync with semantic config?

emodric (approved):
Other than a few nitpicks, :+1:

bundle/NamedObject/Provider/Loading.php
pspanja: Fixed in https://github.com/netgen/ezplatform-site-api/pull/145/commits/b4cee630226e62b70178f91040ffb4b9a2e21b0a

bundle/Resources/config/default_settings.yml
pspanja: Good catch, fixed in https://github.com/netgen/ezplatform-site-api/pull/145/commits/7740fefa397bbe2c0a61d1905c7ab71393db1ba4

pspanja:
Thanks everyone! 👍

PR #146, opened 2019-10-29
[3.5] Move events to bundle

pspanja:
This moves event implementation to bundle. Old event is still dispatched, but event object and class with event constant are deprecated.

Note that I changed naming a bit.

### DIFF
--- a/bundle/EventListener/RenderViewSubscriber.php
+++ b/bundle/EventListener/RenderViewSubscriber.php
@@ -7,7 +7,6 @@ namespace Netgen\Bundle\EzPlatformSiteApiBundle\EventListener;
 use EzSystems\PlatformHttpCacheBundle\ResponseTagger\ResponseTagger;
 use Netgen\Bundle\EzPlatformSiteApiBundle\Event\RenderViewEvent;
 use Netgen\Bundle\EzPlatformSiteApiBundle\Events;
-use Netgen\EzPlatformSiteApi\Event\SiteApiEvents;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
 /**
@@ -28,12 +27,11 @@ final class RenderViewSubscriber implements EventSubscriberInterface
     public static function getSubscribedEvents(): array
     {
         return [
-            Events::RENDER_VIEW => 'tagResponse',
-            SiteApiEvents::RENDER_CONTENT => 'tagResponse',
+            Events::RENDER_VIEW => 'tagView',
         ];
     }
 
-    public function tagResponse(RenderViewEvent $event): void
+    public function tagView(RenderViewEvent $event): void
     {
         $this->responseTagger->tag($event->getView());
     }
--- a/bundle/EventListener/RenderViewSubscriber.php
+++ b/bundle/EventListener/RenderViewSubscriber.php
@@ -7,14 +7,13 @@ namespace Netgen\Bundle\EzPlatformSiteApiBundle\EventListener;
 use EzSystems\PlatformHttpCacheBundle\ResponseTagger\ResponseTagger;
 use Netgen\Bundle\EzPlatformSiteApiBundle\Event\RenderViewEvent;
 use Netgen\Bundle\EzPlatformSiteApiBundle\Events;
-use Netgen\EzPlatformSiteApi\Event\RenderContentEvent;
 use Netgen\EzPlatformSiteApi\Event\SiteApiEvents;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 
 /**
- * @uses \Netgen\Bundle\EzPlatformSiteApiBundle\Events::NG_VIEW_CONTENT_RENDER_VIEW
+ * @uses \Netgen\Bundle\EzPlatformSiteApiBundle\Events::RENDER_VIEW
  */
-final class ViewTaggerListener implements EventSubscriberInterface
+final class RenderViewSubscriber implements EventSubscriberInterface
 {
     /**
      * @var \EzSystems\PlatformHttpCacheBundle\ResponseTagger\ResponseTagger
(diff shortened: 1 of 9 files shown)


bundle/EventListener/ViewTaggerListener.php
```diff
@@ -23,10 +26,10 @@ public function __construct(ResponseTagger $responseTagger)
 
     public static function getSubscribedEvents(): array
     {
-        return [SiteApiEvents::RENDER_CONTENT => 'onRenderContent'];
+        return [Events::NG_VIEW_CONTENT_RENDER => 'onRenderView'];
```
emodric: I feel that this should be kept in sync (as Symfony did in 4.3), e.g.

`return [Events::RENDER_VIEW => 'onRenderView'];`

Btw, the reason why I named it `RenderContentEvent` and `::RENDER_CONTENT` is because you as a developer actually render content and not the view from the Twig templates :)

pspanja: Renamed constant and method in https://github.com/netgen/ezplatform-site-api/pull/146/commits/0f594480db1c0edc7d8c82371f64a6e1302f9f48.

> Btw, the reason why I named it RenderContentEvent and ::RENDER_CONTENT is because you as a developer actually render content and not the view from the Twig templates :)

OK, but the view is what is rendered and the event object can be generic as it's not used for dispatch. So it might make sense to have it this way? :)

emodric: How about just `RENDER_VIEW` then? `NG_VIEW_CONTENT_RENDER` sounds overly complicated :)

pspanja: I renamed it, that's actually `NG_VIEW_CONTENT_RENDER_VIEW` now :))

From my POV it makes sense to have it as descriptive as needed, because that is the actual event that is dispatched, and the Event object is a payload to it. So in this case the event is about rendering a view for `ng_view_content`.

pspanja: As discussed in person, Symfony changed event dispatching in 4.3, so I'm going to update this accordingly. (See https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching)

bundle/EventListener/RenderViewSubscriber.php
```diff
@@ -23,10 +27,13 @@ public function __construct(ResponseTagger $responseTagger)
...
     {
-        return [SiteApiEvents::RENDER_CONTENT => 'onRenderContent'];
+        return [
+            Events::RENDER_VIEW => 'tagResponse',
+            SiteApiEvents::RENDER_CONTENT => 'tagResponse',
+        ];
     }
 
-    public function onRenderContent(RenderContentEvent $event): void
+    public function tagResponse(RenderViewEvent $event): void
```
emodric: Shouldn't the old `RenderContentEvent` extend the new `RenderViewEvent` (currently it is the opposite) to support this?

bundle/Resources/config/services/event_listener.yml
```diff
@@ -1,6 +1,6 @@
 services:
-    netgen.ezplatform_site.event_listener.content_tagger:
-        class: Netgen\Bundle\EzPlatformSiteApiBundle\EventListener\ContentTaggerListener
+    netgen.ezplatform_site.event_listener.render_view:
```
emodric: I feel the name of the event listener should not change here (or rather change only to `ViewTagger`). I mean, there could be more than one RenderView event in the future.

bundle/EventListener/RenderViewSubscriber.php
pspanja: Indeed, but I realized we shouldn't handle both dispatches in the first place, so I'll remove the the deprecated one instead.

emodric: And maybe name the method `tagView`, since that is what you're tagging here.

pspanja: Fixed in https://github.com/netgen/ezplatform-site-api/pull/146/commits/0b248a703dba83c217794063904c3dd29592cb50

emodric: > but I realized we shouldn't handle both dispatches in the first place

Ha, indeed, this is a built in listener after all :D

But the reversal of extending still stands, if someone would implement a listener that works on both 3.x and 4.x.

pspanja: Yup, makes sense. But we're removing it in 4.0, there will be 3.5 for deprecation :)

emodric: I'm aware, but if it's not reversed, you can't really write a listener that will e.g. support 3.4, 3.5 and 4.0. So in `master`, `RenderContentEvent` should definitely extend `RenderViewEvent`, even in future 3.5 version, and on 4.0, just remove `RenderContentEvent`.

pspanja: Fixed extension in https://github.com/netgen/ezplatform-site-api/pull/146/commits/8bec6c2c77266db036d2da2e32fe78cf7bdbf702

bundle/Resources/config/services/event_listener.yml
pspanja: Good point, but in that case I should also change naming of `Events::RENDER_VIEW` to reflect the meaning.

emodric: Are you sure? Event name is one thing, you render a view, hence `Events::RENDER_VIEW`. An event *listener* is the other and it is named based on the purpose in the listener itself.

pspanja: I was thinking that `ViewRenderer`, where the event is dispatched now, is pretty specific, another render view event might happen in another context, for example if we ever need to override view manager from kernel, where the view is also rendered.

emodric: Why should the location of the dispatched event influence the name of the event itself? Why not use the same event at both places? As a consumer of the event, I wouldn't really care where the view render comes from, what I would care only about the fact that a view is rendered.

pspanja: In the view manager example, you would not need to tag the view, since there is already another mechanism that takes care of that, so it's basically a different event. There is a dichotomy here, since you dispatch not by one but by two parameters, event name and event object. Since you subscribe to the event name, that is the actual event, and the event object is just it's payload.

emodric: Hm, okay. Makes sense. In that case, I would deal with it when and *if* we come to it :)

pspanja: I agree, since I don't really have any good naming ideas here :)

Renamed to `ViewTaggerSubscriber` in https://github.com/netgen/ezplatform-site-api/pull/146/commits/ba17c400b002a62ccc450b916855113e79127297

emodric (approved):

pspanja:
Thanks @emodric

PR #147, opened 2019-11-07
NGSTACK-347 enable redirects through content view configuration

iherak:
New feature which enables us to configure redirects directly from the ngcontent_view configuration.

Example usages:
```
ezpublish:
    system:
        frontend_group:
            ngcontent_view:
                full:
                    ng_category:
                        temporary_redirect: "@=content.getFieldRelation('internal_redirect')"
                        match:
                            Identifier\ContentType: ng_category
                    ng_landing_page:
                        temporary_redirect: '@=location.getChildren(1)[0]'
                        match:
                            Identifier\ContentType: ng_landing_page
                  ng_container:
                        permanent_redirect: '@=location.parent'
                        match:
                            Identifier\ContentType: ng_container
                  some_content_type:
                        permanent_redirect: '@=namedObject.getLocation("homepage")'
                        match:
                            Identifier\ContentType: some_content_type
                    other_content_type:
                        redirect:
                            target: "@=namedObject.getTag('running')"
                            target_parameters:
                                foo: bar
                                siteaccess: cro
                            permanent: false
                            absolute: true
                        match:
                            Identifier\ContentType: other_content_type
                    news:
                        redirect:
                            target: 'login'
                            target_parameters:
                                foo: bar
                            permanent: true
                        match:
                            Identifier\ContentType: news
                    blog:
                        redirect:
                            target: 'https://netgen.io'
                        match:
                            Identifier\ContentType: blog
```

### DIFF
--- a/bundle/DependencyInjection/Configuration/Parser/ContentView.php
+++ b/bundle/DependencyInjection/Configuration/Parser/ContentView.php
@@ -39,7 +39,7 @@ class ContentView extends AbstractParser
                                     && !\array_key_exists('redirect', $v);
                             })
                             ->then(function($v) {
-                                $value = \array_key_exists('permanent_redirect', $v) ? $v['permanent_redirect'] : $v['temporary_redirect'];
+                                $value = $v['permanent_redirect'] ?? $v['temporary_redirect'];
                                 $permanent = \array_key_exists('permanent_redirect', $v);
 
                                 $v['redirect'] = [
--- a/bundle/DependencyInjection/Configuration/Parser/ContentView.php
+++ b/bundle/DependencyInjection/Configuration/Parser/ContentView.php
@@ -101,7 +101,7 @@ EOT
                         ->end()
                         ->validate()
                             ->ifTrue(static function ($v): bool {
-                                if  (\array_key_exists('redirect', $v)) {
+                                if (\array_key_exists('redirect', $v)) {
                                     return \array_key_exists('controller', $v) || \array_key_exists('template', $v);
                                 }
 
@@ -113,7 +113,7 @@ EOT
                         ->end()
                         ->validate()
                             ->ifTrue(static function ($v): bool {
-                                if  (\array_key_exists('redirect', $v)) {
+                                if (\array_key_exists('redirect', $v)) {
                                     return \array_key_exists('temporary_redirect', $v) || \array_key_exists('permanent_redirect', $v);
                                 }
 
--- a/bundle/DependencyInjection/Configuration/Parser/ContentView.php
+++ b/bundle/DependencyInjection/Configuration/Parser/ContentView.php
@@ -50,7 +50,7 @@ EOT
                                         ->isRequired()
                                         ->cannotBeEmpty()
                                     ->end()
-                                    ->scalarNode('permanent')
+                                    ->booleanNode('permanent')
                                         ->defaultFalse()
                                     ->end()
                                     ->arrayNode('target_parameters')
--- a/bundle/DependencyInjection/Configuration/Parser/ContentView.php
+++ b/bundle/DependencyInjection/Configuration/Parser/ContentView.php
@@ -35,7 +35,8 @@ class ContentView extends AbstractParser
                     ->arrayPrototype()
                         ->beforeNormalization()
                             ->ifTrue(function($v) {
-                                return \array_key_exists('permanent_redirect', $v) || \array_key_exists('temporary_redirect', $v);
+                                return (\array_key_exists('permanent_redirect', $v) xor \array_key_exists('temporary_redirect', $v))
+                                    && !\array_key_exists('redirect', $v);
                             })
                             ->then(function($v) {
                                 $value = \array_key_exists('permanent_redirect', $v) ? $v['permanent_redirect'] : $v['temporary_redirect'];
@@ -46,11 +47,7 @@ class ContentView extends AbstractParser
                                     'permanent' => $permanent
                                 ];
 
-                                if ($permanent) {
-                                    unset($v['permanent_redirect']);
-                                } else {
-                                    unset($v['temporary_redirect']);
-                                }
+                                unset($v['permanent_redirect'], $v['temporary_redirect']);
 
                                 return $v;
                             })
--- a/bundle/DependencyInjection/Configuration/Parser/ContentView.php
+++ b/bundle/DependencyInjection/Configuration/Parser/ContentView.php
@@ -44,7 +44,8 @@ class ContentView extends AbstractParser
 
                                 $v['redirect'] = [
                                     'target' => $value,
-                                    'permanent' => $permanent
+                                    'permanent' => $permanent,
+                                    'absolute' => false,
                                 ];
 
                                 unset($v['permanent_redirect'], $v['temporary_redirect']);
@@ -70,6 +71,9 @@ EOT
                                         ->cannotBeEmpty()
                                     ->end()
                                     ->booleanNode('permanent')
+                                        ->defaultFalse()
+                                    ->end()
+                                    ->booleanNode('absolute')
                                         ->defaultFalse()
                                     ->end()
                                     ->arrayNode('target_parameters')
--- a/bundle/DependencyInjection/Configuration/Parser/ContentView.php
+++ b/bundle/DependencyInjection/Configuration/Parser/ContentView.php
@@ -64,7 +64,7 @@ EOT
                                 )
                                 ->example('MyBundle:MyControllerClass:view')
                             ->end()
-                            ->arrayNode('redirect') // @todo: add validation
+                            ->arrayNode('redirect')
                                 ->children()
                                     ->scalarNode('target')
                                         ->isRequired()
--- a/bundle/DependencyInjection/Configuration/Parser/ContentView.php
+++ b/bundle/DependencyInjection/Configuration/Parser/ContentView.php
@@ -33,6 +33,28 @@ class ContentView extends AbstractParser
                     ->normalizeKeys(false)
                     ->info("View selection rulesets, grouped by view type. Key is the view type (e.g. 'full', 'line', ...)")
                     ->arrayPrototype()
+                        ->beforeNormalization()
+                            ->ifTrue(function($v) {
+                                return \array_key_exists('permanent_redirect', $v) || \array_key_exists('temporary_redirect', $v);
+                            })
+                            ->then(function($v) {
+                                $value = \array_key_exists('permanent_redirect', $v) ? $v['permanent_redirect'] : $v['temporary_redirect'];
+                                $permanent = \array_key_exists('permanent_redirect', $v);
+
+                                $v['redirect'] = [
+                                    'target' => $value,
+                                    'permanent' => $permanent
+                                ];
+
+                                if ($permanent) {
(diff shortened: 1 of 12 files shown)


bundle/Resources/config/view.yml
```diff
@@ -13,10 +13,24 @@ services:
         arguments:
             - '@netgen.ezplatform_site.ngcontent_view.matcher_factory'
             - '@netgen.ezplatform_site.query_type.query_definition_mapper'
+            - '@netgen.ezplatform_site.redirect_resolver'
         tags:
             - { name: ezpublish.view_provider, type: 'Netgen\Bundle\EzPlatformSiteApiBundle\View\ContentView', priority: 110 }
         public: false
 
+    netgen.ezplatform_site.redirect_resolver:
+        class: Netgen\Bundle\EzPlatformSiteApiBundle\View\Redirect\Resolver
```
emodric, line 22: Why not also `public: false`?

bundle/View/Provider/Configured.php
```diff
@@ -98,16 +109,35 @@ private function getQueriesConfiguration(array $configHash): array
...
-    private function getDTO(array $viewConfig): CoreContentView
+    private function getDTO(array $viewConfig, ContentView $view): CoreContentView
     {
         $dto = new CoreContentView();
         $dto->setConfigHash($viewConfig);
 
+        if (isset($viewConfig['permanent_redirect']) || isset($viewConfig['temporary_redirect'])) {
+            $dto->setControllerReference(
+                new ControllerReference(
+                    'Symfony\Bundle\FrameworkBundle\Controller\RedirectController::urlRedirectAction'
```
emodric: How about `sprintf('%s::%s', RedirectController::class, 'urlRedirectAction')` ?

bundle/View/Redirect/Resolver.php
```diff
@@ -0,0 +1,78 @@
...
+        $this->router = $router;
+    }
+
+    /**
+     * Builds a path to the redirect target.
+     *
+     * Value from configuration can be:
+     *      - SiteAPI location
+     *      - SiteAPI content
+     *      - location id fetched from config resolver
```
emodric: content ID could be added also, which can generate a link to a main location, for example, if you extract values from relation (list) fields.

bundle/View/Redirect/Resolver.php
```diff
@@ -0,0 +1,78 @@
...
+    public function resolveTarget(string $redirectConfig, ContentView $view): string
+    {
+        $value = $this->parameterProcessor->process($redirectConfig, $view);
+
+        if (is_array($value)) {
+            if (count($value) < 1) {
+                return '/';
+            }
+
+            $value = reset($value);
```
emodric: `array_key_first`, together with https://github.com/symfony/polyfill-php73?

EDIT: Disregard, you're fetching the first value, not key :)

bundle/View/Redirect/ParameterProcessor.php
```diff
@@ -0,0 +1,99 @@
...
+
+        $this->registerFunctions($language);
+
+        return $language->evaluate(
+            \substr($value, 2),
+            [
+                'location' => $view->getSiteLocation(),
+                'content' => $view->getSiteContent(),
+                'configResolver' => $this->configResolver,
+                'namedObject' => $this->namedObjectProvider,
```
emodric, line 49: `namedObjects` maybe? since you will use it with e.g. `namedObjects.location(42)`.

bundle/View/Redirect/ParameterProcessor.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /**
+     * Register functions with the given $expressionLanguage.
+     *
+     * @param \Symfony\Component\ExpressionLanguage\ExpressionLanguage $expressionLanguage
+     */
+    private function registerFunctions(ExpressionLanguage $expressionLanguage): void
+    {
+        $configResolver = $this->configResolver;
```
emodric: You don't need this. You can use non static closures, which are bound to `$this` by default.

bundle/View/Redirect/ParameterProcessor.php
iherak: This was used same as the existing configuration for named objects

emodric:
How about adding support for Symfony routes and Netgen Tags?

bundle/View/Redirect/Resolver.php
iherak: Not sure what you mean here? How would that configuration look like?
I'm talking here about this:
`temporary_redirect: "@=configResolver.getParameter("some_location_id", "ngsite")"`

Maybe rewording of the docs is needed here?

emodric:
Thinking about future improvements, maybe it would be good to have a config similar to this:

```
ezpublish:
    system:
        frontend_group:
            ngcontent_view:
                full:
                    ng_category:
                        redirect:
                            target: "@=content.getFieldRelation('internal_redirect')"
                            # Symfony route
                            # target: "ngsite_content_search"
                            # When using a route, you can use these parameters to generate the route
                            target_parameters:
                                foo: bar
                                bar: baz
                                location: "@=configResolver.parameter('foo', 'bar')"
                            permanent: false
                        match:
                            Identifier\ContentType: ng_category
```

We could keep `temporary_redirect` and `permanent_redirect` as shortcuts, too.

bundle/View/Redirect/Resolver.php
emodric: What I meant is, I'd definitelly expect to be able to also use a content ID for redirection too. There is no place where it's documented that `temporary_redirect: "@=configResolver.getParameter("some_location_id", "ngsite")"` works, but `temporary_redirect: "@=configResolver.getParameter("some_content_id", "ngsite")"` does not work.

The only place where that is apparent is in the resolver using the hardcoded `locationId` parameter to `ez_url_alias` route.

emodric: Besides, you don't have to use config resolver to provide the ID. You could just as well use `temporary_redirect: 42` and it would work. Is it a location ID or a content ID?

emodric: It was internally agreed to just remove support for IDs as targets, too much ambiguity to begin with.

bundle/View/Redirect/ParameterProcessor.php
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    /**
+     * Register functions with the given $expressionLanguage.
+     *
+     * @param \Symfony\Component\ExpressionLanguage\ExpressionLanguage $expressionLanguage
+     */
+    private function registerFunctions(ExpressionLanguage $expressionLanguage): void
+    {
+        $namedObjectProvider = $this->namedObjectProvider;
```
emodric: The comment about non-static closures and `$this` applies to this too :)

bundle/View/Redirect/Resolver.php
```diff
@@ -0,0 +1,69 @@
...
+     * @param string $redirectConfig
+     * @param \Netgen\Bundle\EzPlatformSiteApiBundle\View\ContentView $view
+     *
+     * @return string
+     */
+    public function resolveTarget(string $redirectConfig, ContentView $view): string
+    {
+        $value = $this->parameterProcessor->process($redirectConfig, $view);
+
+        if (is_array($value)) {
```
emodric: In what cases is this value an array?

iherak: for example, `@=location.getChildren(1)` in case we want to redirect to first child.

emodric: Can't that be written as `@=location.getChildren(1)[0]` ? It seems weird supporting arrays for such a trivial usecase, especially when you only use the first value. Kinda implies that if you use only a first value of an array, you can just as well transfer the value itself.

iherak: Damn, that works. Will remove!

iherak:
> What's the reasoning for making the temporary redirect default? Not leaning on either, I'm just interested in terms of docs.

It makes more sense to me. Easier to change from temporary to permanent than vice-versa. Browsers for example cache permanent redirects quite heavily, so this might make it easier on the developers life.

bundle/DependencyInjection/Configuration/Parser/ContentView.php
```diff
@@ -68,6 +99,50 @@ public function addSemanticConfig(NodeBuilder $nodeBuilder): void
                                 ->variablePrototype()->end()
                             ->end()
                         ->end()
+                        ->validate()
```
pspanja, line 125: It should be possible to merge this with validation below, handling the same for `temporary_redirect` and `permanent_redirect`.

bundle/View/Redirect/RedirectConfiguration.php
```diff
@@ -0,0 +1,70 @@
...
+     * @param bool $permanent
+     */
+    public function __construct(string $target, array $targetParameters, bool $permanent)
+    {
+        $this->target = $target;
+        $this->targetParameters = $targetParameters;
+        $this->permanent = $permanent;
+    }
+
+    public static function fromConfigurationArray($config): RedirectConfiguration
```
pspanja: Typehint `$config` as an array.

bundle/View/Redirect/Resolver.php
```diff
@@ -0,0 +1,70 @@
...
+        );
+
+        if ($value instanceof Location || $value instanceof Content) {
+            return $this->router->generate(
+                $value,
+                $redirectConfig->getTargetParameters()
+            );
+        }
+
+        return '/';
```
pspanja: This means invalid configuration, should we rather throw an exception/log a critical message?

bundle/View/Redirect/ParameterProcessor.php
iherak: Fixed now :)

iherak:
Ok, I think all remarks have been addressed, support for Symfony routes and Netgen Tags has been added. Please do a new round of reviews :)

bundle/View/Redirect/Resolver.php
```diff
@@ -84,11 +84,7 @@ private function processUrl(RedirectConfiguration $redirectConfig): string
...
         $url = $redirectConfig->getTarget();
 
-        $queryStringArray = [];
-        foreach ($redirectConfig->getTargetParameters() as $key => $value) {
-            $queryStringArray[] = \urlencode($key) . '=' . \urlencode($value);
-        }
-
-        return \count($queryStringArray) > 0 ? $url . '?' . \implode('&', $queryStringArray) : $url;
+        return \count($redirectConfig->getTargetParameters()) > 0 ?
+            $url . '?' . \http_build_query($redirectConfig->getTargetParameters()) : $url;
```
emodric: Final nitpick, I promise :)

What if the URL already has query params?

emodric: After internal discussion, it was decided to simply skip using `target_parameters` in case an absolute URL was provided. Absolute URL can contain query parameters and/or a fragment, which would only complicate adding custom query params taken from `target_parameters`.

emodric (approved):

emodric:
Nice work @iherak !

pspanja (changes requested):

pspanja (approved):

pspanja:
Thanks @iherak, great work!

PR #148, opened 2019-11-19
[3.3] Add Location::getFirstChild() method

pspanja:
Lately I've seen some uses of getting the first child of a Location, implemented through the `getChildren()` method:

```twig
{{ location.children(1)[0].content.name }}
```

Method `getChildren()` returns an array of Location objects, intended for iteration, and using it like that is awkward and does not provide for limiting on the ContentType. For that reason I'm adding a new method on the Location object:

```php
Location::getFirstChild(?string $contentTypeIdentifier = null): ?Location
```

Use from Twig:

```twig
{% set firstChild = location.firstChild('article') %}

{% if firstChild is not null %}
    {{ firstChild.content.name }}
{% endif %}
```

Or in redirect configuration:

```yaml
ezpublish:
  system:
    frontend_group:
      ngcontent_view:
        full:
          ng_frontpage:
            temporary_redirect: '@=location.getFirstChild'
            match:
              Identifier\ContentType: ng_frontpage
```

### DIFF
--- a/lib/API/Values/Location.php
+++ b/lib/API/Values/Location.php
@@ -58,7 +58,7 @@ abstract class Location extends ValueObject
     /**
      * Return first child, limited by optional $contentTypeIdentifier.
      *
-     * @param string|null $contentTypeIdentifier
+     * @param null|string $contentTypeIdentifier
      *
      * @return null|\Netgen\EzPlatformSiteApi\API\Values\Location
      */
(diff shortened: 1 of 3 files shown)


emodric (approved):

MarioBlazek (approved):

pspanja:
Thanks!

PR #149, opened 2019-11-26
Fix preview of the unpublished Content

pspanja:
Generating virtual Location for the unpublished Content preview did not work, because Location constructor was changed since the controller implementation.


emodric (approved):

pspanja:
Thanks @emodric

PR #150, opened 2019-12-01
[3.3] NGSTACK-363: Implement Content::getFirstNonEmptyField()

pspanja:
This implements a new method on `Content` and `Field` objects:

```php
public function getFirstNonEmptyField(
    string $firstIdentifier, 
    string ...$otherIdentifiers
);
```

The method will return first existing and non-empty Field from the given Field identifiers.
If none of the fields are found or all found fields are empty, a `ngsurrogate` type field will be returned, which renders to an empty string.

In case this method is used in Twig to choose among Fields of multiple types, accessing the value directly would be ambiguous without first checking for the type of the returned Field. In that case it's best to use the method together with `ng_render_field`:

```twig
{{ ng_render_field(content.firstNonEmptyField('title', 'short_title', 'name')) }}
```

### DIFF
--- a/lib/API/Values/Content.php
+++ b/lib/API/Values/Content.php
@@ -78,7 +78,7 @@ abstract class Content extends ValueObject
      *
      * @return \Netgen\EzPlatformSiteApi\API\Values\Field
      */
-    abstract public function getFirstSetField(string $firstIdentifier, string ...$otherIdentifiers): Field;
+    abstract public function getFirstNonEmptyField(string $firstIdentifier, string ...$otherIdentifiers): Field;
 
     /**
      * Returns a field value for the given field definition identifier.
--- a/lib/API/Values/Content.php
+++ b/lib/API/Values/Content.php
@@ -68,6 +68,19 @@ abstract class Content extends ValueObject
     abstract public function getFieldById($id): Field;
 
     /**
+     * Return the first existing and non-empty field.
+     *
+     * If no field is found in the Content, a surrogate field will be returned.
+     * If all found fields are empty, the first found field will be returned.
+     *
+     * @param string $firstIdentifier
+     * @param string ...$otherIdentifiers
+     *
+     * @return \Netgen\EzPlatformSiteApi\API\Values\Field
+     */
+    abstract public function getFirstSetField(string $firstIdentifier, string ...$otherIdentifiers): Field;
+
+    /**
      * Returns a field value for the given field definition identifier.
      *
      * @param string $identifier
--- a/lib/Core/Site/Values/Fields.php
+++ b/lib/Core/Site/Values/Fields.php
@@ -237,6 +237,45 @@ final class Fields extends APIFields
     }
 
     /**
+     * {@inheritDoc}
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     */
+    public function getFirstSetField(string $firstIdentifier,  string ...$otherIdentifiers): APIField
+    {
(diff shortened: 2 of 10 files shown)


lib/API/Values/Content.php
```diff
@@ -67,6 +67,19 @@ abstract class Content extends ValueObject
...
+     *
+     * If no field is found in the Content, a surrogate field will be returned.
+     * If all found fields are empty, the first found field will be returned.
+     *
+     * @param string $firstIdentifier
+     * @param string ...$otherIdentifiers
+     *
+     * @return \Netgen\EzPlatformSiteApi\API\Values\Field
+     */
+    abstract public function getFirstSetField(string $firstIdentifier, string ...$otherIdentifiers): Field;
```
emodric: Why not just `getFirstSetField(string ...$identifiers)` ? To ensure that at least one identifier is passed?

lib/Core/Site/Values/Fields.php
```diff
@@ -236,6 +236,45 @@ public function getFieldById($id): APIField
...
+     *
+     * @throws \eZ\Publish\API\Repository\Exceptions\NotFoundException
+     */
+    public function getFirstSetField(string $firstIdentifier, string ...$otherIdentifiers): APIField
+    {
+        $identifiers = \array_merge([$firstIdentifier], $otherIdentifiers);
+        $fields = $this->getAvailableFields($identifiers);
+
+        foreach ($fields as $field) {
+            if (!$field->isEmpty()) {
```
emodric, line 250: If we're checking for emptiness, should be call the method `getFirstNonEmptyField` or something similar?

"first set" field could be interpreted as the first field that exists in the content object.

lib/API/Values/Content.php
pspanja: Exactly, otherwise I would have to validate the input manually.

lib/Core/Site/Values/Fields.php
```diff
@@ -236,6 +236,45 @@ public function getFieldById($id): APIField
...
+                return $field;
+            }
+        }
+
+        return $fields[0] ?? $this->getSurrogateField($firstIdentifier, $this->content);
```
emodric, line 255: Why do we need `$fields[0]` here?

So, if we have 5 fields, we return the first non-empty one, but if there are no empty ones, we return first field even if it is empty. Sounds too much complicated.

lib/Core/Site/Values/Fields.php
pspanja: Makes sense, I'll update.

lib/Core/Site/Values/Fields.php
pspanja: The reason is I wanted to return surrogate field only if no fields were found. But I agree, it is a bit complicated.

lib/Core/Site/Values/Fields.php
pspanja: Changed in https://github.com/netgen/ezplatform-site-api/pull/150/commits/f54d285ef778df22632e76347106d1c49d361b45

lib/Core/Site/Values/Fields.php
pspanja: Changed in https://github.com/netgen/ezplatform-site-api/pull/150/commits/c76c33d49e6c1655b6b990ef7899cc9d549fcc42

pspanja: Restored returning first empty field, as it might be useful if for example using field definition in some way.

pspanja:
Thanks for the review!

PR #152, opened 2019-12-09
[3.3] NGSTACK-371: implement fallback between eZ and Site API content views

pspanja:
This implements fallback between Site API and eZ Platform content views using templates that render secondary content view. Secondary content view can be eZ Platform or Site API, depending on Site API content view being enabled through `override_url_alias_view_action`.

This will significantly lower the bar when introducing Site API to an existing project, as it enables seamless use of both content views together, making having a functional site from the beginning and incremental migration to Site API possible. For a contingency, the opposite also works and one can easily introduce eZ Platform content view fallback into an existing Site API project.

Fallback is controlled through two configuration options, available under `ezpublish/system` node (showing defaults):

- `ng_fallback_to_secondary_content_view: false`
  enables content view fallback, should become `true` by default in the next major release
- `ng_fallback_with_subrequest: true`
  controls whether fallback should use a subrequest

To enable fallback to eZ Platform content view without using a subrequest, a new Twig function is implemented:

- `ng_ez_view_content(value, viewType, parameters, layout)`

This function is symmetrical to the existing `ng_view_content` function.

Since the fallback is achieved though templates, it is possible to manually configure fallback only for specific views by configuring a fallback template, for example using a match-all view for eZ Platform content view:

```yaml
content_view:
    full:
        default:
            template: "@NetgenEzPlatformSiteApi/content_view_fallback/to_site_api.html.twig"
            match: ~
```

Or similar for Site API content view:

```yaml
ngcontent_view:
    full:
        default:
            template: "@NetgenEzPlatformSiteApi/content_view_fallback/to_ez_platform.html.twig"
            match: ~
```

### Some notes about implementation
- logic for rendering a `View` using and controller without a subrequest is extracted into `ViewRenderer` service
- checking for default controller is simplified and checks only for controller not being configured, as the default controller only return the given view anyway
- setting default values for `ezpublish` semantic configuration from PHP in the bundle extension is replaced with using `addDefaultSettings()` method on `Core` bundle extension from eZ Platform kernel
- view services are no longer loaded separately in the bundle extension, instead `view.yml` is imported from the main `services.yml`

### To dos
- [x] write documentation


bundle/Resources/views/content_view_fallback/to_ez_platform.html.twig
```diff
@@ -0,0 +1,13 @@
...
+        controller(
+            'ez_content:viewAction', {
+                'content': content.innerContent,
+                'location': location.innerLocation,
+                'viewType': view_type
+            }
+        )
+    ) }}
+{% else %}
+    {{ ng_ez_view_content(location|default(content), view_type) }}
```
emodric: content.innerContent & location.innerLocation ?

bundle/DependencyInjection/NetgenEzPlatformSiteApiExtension.php
```diff
@@ -78,22 +75,4 @@ public function prepend(ContainerBuilder $container): void
...
     }
-
-    /**
-     * Default values must be set conditionally because eZ Kernel bundles are usually activated
-     * before this extension is ran. In that case the values that were already processed and merged
-     * to the container would be overwritten with the default value.
-     *
-     * @param \Symfony\Component\DependencyInjection\ContainerBuilder $container
-     */
-    private function setDefaultValuesIfNeeded(ContainerBuilder $container): void
```
emodric, line 89: Will this change make sure that we do not get back the issue described in the comment here?

pspanja: It should, this is what eZ's bundles use. But TBH I didn't test it, will try to reproduce it.

bundle/Resources/views/content_view_fallback/to_ez_platform.html.twig
pspanja: `ng_ez_view_content` also handles Site API Content and Location, but using inner objects would be more clear, will update.

bundle/DependencyInjection/NetgenEzPlatformSiteApiExtension.php
pspanja: Reproduced and validated, this fixes the problem.

bundle/Resources/views/content_view_fallback/to_ez_platform.html.twig
pspanja: Updated in https://github.com/netgen/ezplatform-site-api/pull/152/commits/66c10861232d12fd8ea4d8517a2e7bea6a980081.

pspanja:
Thanks @emodric

PR #153, opened 2019-12-22
Implement Content View Location resolver

pspanja:
This implements `LocationResolver` for the Content View, aligning the implementation with changes in kernel introduced in https://github.com/ezsystems/ezpublish-kernel/pull/1802.

We implement our own Location resolver (loader), to be more flexible in the future when introducing cross siteaccess linking. Adapter for the `ContentInfoLocationLoader` from kernel is also provided.

With this change, loading the Location in `ng_view_content` and `ng_ez_view_content` is removed, as it's now done by the view builder.

Additionally fixed using `ng_content::viewAction` instead of `ez_content::viewAction` in `ng_ez_view_content`.

### DIFF
--- a/bundle/Resources/config/services/templating.yml
+++ b/bundle/Resources/config/services/templating.yml
@@ -27,7 +27,6 @@ services:
         arguments:
             - '@netgen.ezplatform_site.view_builder.content'
             - '@netgen.ezplatform_site.view_renderer'
-            - '@ezpublish.api.service.location'
         tags:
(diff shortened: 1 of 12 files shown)


pspanja:
@hknezevic FYI, couple of days ago we discussed cross-siteaccess linking in single repo multisite installations, this is one piece we'll need to implement proper handling of content embeds. For example `SiteaccessSubtreeLocationProvider` or something like that.

PR #154, opened 2019-12-23
ezp v3 support: missing argument from EZP-30882

wizhippo:
https://github.com/ezsystems/ezpublish-kernel/pull/2868


pspanja:
Thanks @wizhippo!

PR #155, opened 2019-12-23
NGSTACK-354: implement support for XmlText and RichText embed rendering

pspanja:
This implements support for XmlText and RichText field types embed rendering.

Two semantic configuration options are provided to control if the embeds are rendered with or without a subrequest (showing defaults):

`ng_richtext_embed_without_subrequest: false`
`ng_xmltext_embed_without_subrequest: false`

Both will be set to true by default in the next major release.
To enable rendering RichText embeds without a subrequest a new Twig functions `ng_view_content_embedded` and `ng_ez_view_content_embedded` are implemented. This functions works with Content and Location IDs, since Content and Location objects are not provided to the embed templates.

For those already using Content View fallback, note that the configuration option to control rendering without a subrequest is here changed from `ng_fallback_with_subrequest: true` to `ng_fallback_without_subrequest: false`, as it makes more sense to have it true when in becomes the default in the future.


bundle/Resources/views/content_view_fallback/to_ez_platform/embed.html.twig
```diff
@@ -0,0 +1,16 @@
+{# content \Netgen\EzPlatformSiteApi\API\Values\Content #}
+{# location \Netgen\EzPlatformSiteApi\API\Values\Location #}
+
+{% if ezpublish.configResolver.getParameter('ng_fallback_without_subrequest') == true %}
+    {{ ng_ez_view_content_embedded(view_type, { "content": content.innerContent, "location": location.innerLocation }) }}
+{% else %}
+    with subrequest {{ render(
```
emodric: Is `with subrequest` here a typo?

bundle/Resources/views/content_view_fallback/to_site_api/embed.html.twig
```diff
@@ -0,0 +1,16 @@
+{# content \Netgen\EzPlatformSiteApi\API\Values\Content #}
+{# location \Netgen\EzPlatformSiteApi\API\Values\Location #}
+
+{% if ezpublish.configResolver.getParameter('ng_fallback_without_subrequest') == true %}
+    {{ ng_view_content_embedded(view_type, { "content": content, "location": location }) }}
```
emodric, line 5: Isn't it `ng_ez_view_content_embedded` ? Multiple instances exist.

emodric: Ah, different methods, for site API and eZ.

emodric (approved):

bundle/Resources/views/content_view_fallback/to_ez_platform/embed.html.twig
pspanja: A leftover from testing :) Removed in https://github.com/netgen/ezplatform-site-api/pull/155/commits/e6465d97407f9b5af02e3038859c98dce775f309

pspanja:
Thanks @emodric

PR #157, opened 2020-01-30
NGSTACK-389: implement view configuration inheritance

pspanja:
This implements view configuration inheritance through the `extends` key, allowing rewrite of the following configuration:

```yaml
standard:
    common:
        template: "@ezdesign/content/standard/{content_type}.html.twig"
        params:
            with_intro: false
        match:
            Identifier\ContentType:
                - file
                - image
                - ng_article
                - ng_audio
                - ng_banner
                - ng_blog_post
                - ng_category
                - ng_feedback_form
                - ng_frontpage
                - ng_gallery
                - ng_htmlbox
                - ng_landing_page
                - ng_news
                - ng_recipe
                - ng_video
    match_all:
        template: "@ezdesign/content/standard.html.twig"
        params:
            with_intro: false
        match: ~
standard_with_intro:
    common:
        template: "@ezdesign/content/standard/{content_type}.html.twig"
        params:
            with_intro: true
        match:
            Identifier\ContentType:
                - file
                - image
                - ng_article
                - ng_audio
                - ng_banner
                - ng_blog_post
                - ng_category
                - ng_feedback_form
                - ng_frontpage
                - ng_gallery
                - ng_htmlbox
                - ng_landing_page
                - ng_news
                - ng_recipe
                - ng_video
    match_all:
        template: "@ezdesign/content/standard.html.twig"
        params:
            with_intro: true
        match: ~
```

To this:

```yaml
standard:
    common:
        template: "@ezdesign/content/standard/{content_type}.html.twig"
        params:
            with_intro: false
        match:
            Identifier\ContentType:
                - file
                - image
                - ng_article
                - ng_audio
                - ng_banner
                - ng_blog_post
                - ng_category
                - ng_feedback_form
                - ng_frontpage
                - ng_gallery
                - ng_htmlbox
                - ng_landing_page
                - ng_news
                - ng_recipe
                - ng_video
    match_all:
        template: "@ezdesign/content/standard.html.twig"
        params:
            with_intro: false
        match: ~
standard_with_intro:
    common:
        extends: standard/common
        params:
            with_intro: true
    match_all:
        extends: standard/match_all
        params:
            with_intro: true
```

Inheritance is limited to one level only. Also, because of how the configuration processing is implemented, the extended configuration must be defined under the same `ngcontent_view` key. I think this is not a problem and will actually help in keeping this feature under control, avoiding inheritance through multiple scopes that might be hard to understand at a glance.

Note that when merging with the extended configuration, only first level keys are used. That means for example you can't override individual parameters, they are either defined in full or inherited from the extended configuration.

Validation of the `match` not being defined when `extends` is used had to be implemented in the before-normalization phase, as the usage of prototype under the array definition will create an empty array value even if the node is marked as required. In that case, required relates to the provided value, not to the resolved value available after processing.

### DIFF
--- a/bundle/DependencyInjection/Configuration/Parser/ContentView.php
+++ b/bundle/DependencyInjection/Configuration/Parser/ContentView.php
@@ -8,6 +8,7 @@ use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\AbstractPars
 use eZ\Bundle\EzPublishCoreBundle\DependencyInjection\Configuration\SiteAccessAware\ContextualizerInterface;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\NodeBuilder;
+use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Twig\Lexer;
 
 class ContentView extends AbstractParser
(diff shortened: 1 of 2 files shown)


zulicek-dumancic (approved):

Ljudevit (approved):
Awesome feature!

pspanja:
Thanks

PR #159, opened 2020-03-31
NGSTACK-410: fix extended config not overriding default values

pspanja:
View configuration sets some default values (for `match`, `queries` and `params` keys) that are not being overridden with extended config values when using `extends` option.


ludwig031 (approved):
Tested and everything works now

pspanja:
Thanks @ludwig031

PR #161, opened 2020-05-15
NGSTACK-413 Check if field relation value is empty

RandyCupic:
When fetching related content in Site API from single field relation, if there's no related content, it throws an error `Unsupported value (NULL)` because it creates a `ContentId` criterion with `null`.


pspanja (approved):

pspanja:
Thanks @RandyCupic!

PR #163, opened 2020-07-03
Handle exception when loading a named location

leohajder:
`{% set location = ng_named_location('location_name') %}` throws an exception if the user has no access permission to view the location, and there is no way of handling it in a twig template.

### DIFF
--- a/bundle/Templating/Twig/Extension/NamedObjectRuntime.php
+++ b/bundle/Templating/Twig/Extension/NamedObjectRuntime.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Netgen\Bundle\EzPlatformSiteApiBundle\Templating\Twig\Extension;
 
+use eZ\Publish\Core\Base\Exceptions\UnauthorizedException;
 use Netgen\Bundle\EzPlatformSiteApiBundle\NamedObject\Provider;
 use Netgen\EzPlatformSiteApi\API\Values\Content;
 use Netgen\EzPlatformSiteApi\API\Values\Location;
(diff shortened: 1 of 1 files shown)


emodric:
Should we log the exception to an error log (with `critical` type)?

bundle/Templating/Twig/Extension/NamedObjectRuntime.php
```diff
@@ -35,13 +36,17 @@ public function getNamedContent(string $name): ?Content
...
+        public function getNamedLocation(string $name): ?Location
     {
-        if ($this->namedObjectProvider->hasLocation($name)) {
-            return $this->namedObjectProvider->getLocation($name);
+        try {
+            if ($this->namedObjectProvider->hasLocation($name)) {
+                return $this->namedObjectProvider->getLocation($name);
+            }
+
+            return null;
```
emodric, line 46: You can just have one `return null;` after the `try/catch` block.

ilukac:
> Should we log the exception to an error log (with `critical` type)?

I don't think it is critical. Could be a location which not all users have read perm.

bundle/Templating/Twig/Extension/NamedObjectRuntime.php
leohajder: I know. I'm working on a new PR to handle exceptions in all methods in `NamedObjectRuntime`, and I have already refactored this. I will close this PR later.

emodric:
The problem is, nothing but `critical` goes into log in production. If it goes to Sentry, then it's not good, that's for sure.

emodric:
We can log only if in debug mode, too, to prevent it going to Sentry in production.

bundle/Templating/Twig/Extension/NamedObjectRuntime.php
leohajder: I have opened a new PR: https://github.com/netgen/ezplatform-site-api/pull/164
Closing this...

leohajder:
Closed, handled by https://github.com/netgen/ezplatform-site-api/pull/164

PR #165, opened 2020-07-15
Enable autowired controllers

MarioBlazek:
This PR allows controllers to be **autowired**, which means that explicit service definition for a controller is not required anymore. By default SiteAPI controller extends the **AbstractController** which implements **ServiceSubscriberInterface** with the set of necessary services. We need to expand **getSubscribedServices()** method with services that are necessary for the Site API controller.

The use case we want to achieve. Custom view controller without service definition:

```yaml
ng_recipe:
    template: "@ezdesign/content/full/ng_recipe.html.twig"
    controller: App\Controller\ExampleController::view
    match:
        Identifier\ContentType: ng_recipe
```

```php
<?php

namespace App\Controller;

use Netgen\Bundle\EzPlatformSiteApiBundle\Controller\Controller;
use Netgen\Bundle\EzPlatformSiteApiBundle\View\ContentView;

class ExampleController extends Controller
{
    public function view(ContentView $view)
    {
        $this->getConfigResolver();
        $this->getGlobalHelper();
        $this->getQueryTypeRegistry();
        $this->getRepository();

        $this->getNamedObjectProvider();
        $this->getSite();

        return $view;
    }
}
```

Without this code, calling **$this->getSite()** would result in this error:
```
Service "netgen.ezplatform_site.site" not found: even though it exists in the app's container, the container inside "App\Controller\ExampleController" is a smaller service locator that only knows about the "doctrine", "form.factory", "http_kernel", "parameter_bag", "request_stack", "router", "security.authorization_checker", "security.csrf.token_manager", "security.token_storage", "serializer", "session" and "twig" services. Try using dependency injection instead.
```

### DIFF
--- a/bundle/Controller/Controller.php
+++ b/bundle/Controller/Controller.php
@@ -100,4 +100,16 @@ abstract class Controller extends AbstractController
 
         return $namedObjectProvider;
     }
+
+    public static function getSubscribedServices()
+    {
+        return [
+            'netgen.ezplatform_site.site' => Site::class,
+            'netgen.ezplatform_site.named_object_provider' => Provider::class,
+            'ezpublish.query_type.registry' => QueryTypeRegistry::class,
+            'ezpublish.api.repository' => Repository::class,
+            'ezpublish.templating.global_helper' => GlobalHelper::class,
+            'ezpublish.config.resolver' => ConfigResolverInterface::class,
+        ] + parent::getSubscribedServices();
+    }
 }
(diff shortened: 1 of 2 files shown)


emodric (approved):
Are the service definitions necessary? Can we just directly reference the service names in `getSubscribedServices()` ?

Looks good! :+1:

MarioBlazek:
Service definitions are necessary because Symfony requires the value of array to be proper PHP type.

This won't work:

```php
return [
    'service_id' => 'service_id',
];
```

emodric:
Alright :+1:

emodric:
Btw, don't we need an entry in `NetgenEzPlatformSiteApiExtension` to load `autowiring.yml`?

MarioBlazek:
No, the main service.yml file has:

```yml
imports:
    - { resource: services/*.yml }
```

emodric:
Ok, I'm apparently blind :D I was looking at the file and couldn't see it !

pspanja (approved):

pspanja:
Awesome, thanks @MarioBlazek!

PR #170, opened 2020-09-01
NGSTACK-411 Catch sort clause not implemented exception

RandyCupic:
Methods such as `filterChildren()` on Location internally use `getSortClausses()` method on parent location to determine sorting. If a sorting option, that is not implemented (eg. `Class name` in new stack), is selected in admin, it will throw an exception which is being propagated all the way up and it crashes the site.

This PR wraps the call for getting location's sort clauses into the try-catch block and uses empty array if sort clause is not implemented.

@pspanja Would it make sense to use some other default sort clause instead of empty array? Should I log this exception?


pspanja (approved):
Thanks @RandyCupic!

PR #172, opened 2020-09-07
NGSTACK-416: set layout to true for full view fallback

pspanja:
The check is hardcoded in the template, as the `full` literal is not likely to change.

### DIFF
--- a/bundle/Resources/views/content_view_fallback/to_ez_platform/view.html.twig
+++ b/bundle/Resources/views/content_view_fallback/to_ez_platform/view.html.twig
@@ -1,15 +1,19 @@
 {# content \Netgen\EzPlatformSiteApi\API\Values\Content #}
 {# location \Netgen\EzPlatformSiteApi\API\Values\Location #}
 
+{% set viewType = app.request.attributes.get('viewType') %}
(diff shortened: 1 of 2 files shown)


RandyCupic (approved):

RandyCupic:
@pspanja Tested on a project (Kavli) where the issue has been discovered, works fine.

emodric:
@RandyCupic What was the issue?

pspanja:
@emodric the problem was when using `noLayout` variable that is injected in the template, for example:

```twig

{% set pagelayout = viewbaseLayout is defined ? viewbaseLayout : 'NetgenSiteBundle::pagelayout_empty.html.twig' %}

{% if (noLayout is not defined or not noLayout) and ezpublish.configResolver.hasParameter('pagelayout', 'ngsite') %}
    {% set pagelayout = ezpublish.configResolver.getParameter('pagelayout', 'ngsite') %}
{% endif %}

{% extends pagelayout %}
...
```

In that case layout would not be set correctly as `noLayout` is `true` by default.

emodric (approved):

pspanja:
Thanks!

PR #173, opened 2020-09-07
NGSTACK-429: add methods to RelationService that work with the given Content version

pspanja:
This changes methods of the `RelationService`:

- `loadFieldRelation(Content $content, string $fieldDefinitionIdentifier, array $contentTypeIdentifiers = [])`
- `loadFieldRelations(Content $content, string $fieldDefinitionIdentifier, array $contentTypeIdentifiers = [])`

Using ID as the first argument still works, but is now deprecated.

The goal is to get relations of a given `Content` version, needed in the preview.

### DIFF
--- a/docs/reference/services.rst
+++ b/docs/reference/services.rst
@@ -237,55 +237,6 @@ Methods
     :local:
 
 ``loadFieldRelation()``
-.......................
-
-Load single field relation from a specific field in a published version of a specific Content.
-
-The method will return ``null`` if the field does not contain relations that can be loaded by the
-current user. If the field contains multiple relations, the first one will be returned. The method
-supports optional filtering by ContentType.
-
-+----------------------------------------+------------------------------------------------------------------------------------+
-| **Parameters**                         | 1. ``string|int $contentId``                                                       |
-|                                        | 2. ``string $fieldDefinitionIdentifier``                                           |
-|                                        | 3. ``array $contentTypeIdentifiers = []``                                          |
-+----------------------------------------+------------------------------------------------------------------------------------+
-| **Returns**                            | :ref:`Content object<content_object>` or ``null``                                  |
-+----------------------------------------+------------------------------------------------------------------------------------+
-| **Example**                            | .. code-block:: php                                                                |
-|                                        |                                                                                    |
-|                                        |     $content = $relationService->loadFieldRelation(                                |
-|                                        |         42,                                                                        |
-|                                        |         'relations',                                                               |
-|                                        |         ['articles']                                                               |
-|                                        |     );                                                                             |
-|                                        |                                                                                    |
-+----------------------------------------+------------------------------------------------------------------------------------+
-
-``loadFieldRelations()``
-........................
-
-Load all field relations from a specific field in a published version of a specific Content. The
-method supports optional filtering by ContentType.
-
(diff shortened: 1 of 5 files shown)


pspanja:
Q: do we need `RelationService::load*` methods or they can be deprecated and removed?

RandyCupic (approved):

RandyCupic:
@pspanja Tested on project (Kavli) where it was reported from, works fine.

emodric:
Why would we even need additional methods?

Just deprecate content ID in 3.5 and replace it with real content in 4.0 ?

iherak:
I presume it's about backporting - and if that's the case, I am fine with adding new methods in 3.5 and deprecating old ones, and removing load* methods in 4.0

emodric:
@iherak How about keeping the method names, just repurposing the first argument from content ID to actual content (with proper BC policy, ofcourse)?

iherak:
Well, that would be a BC break, right? So I'd vote against it, realistically, few people read release notes when upgrading minor version.

emodric:
Sure, but that's why there would be a deprecation message in 3.5 AND a BC layer accepting both ID and content. We already broke 4.0 with config changes and so on, so I don't see it as an issue.

From the point of view of a developer, what's the difference between having to use a new method instead of the old one, and having to provide a content to existing method instead of an ID?

iherak:
Ah, with a BC layer in 3.5 it would be perfectly ok. So with that :+1: from me.

And for your second question - it would be a difference, as old methods still work for most use-cases, it wouldn't break anything. You would use new methods as a way to improve the way it works, but if you don't, nothing should break.

emodric:
@iherak 

Well, not with this comment from you they don't :D

> and removing load* methods in 4.0

To be clear, I was never suggesting doing anything without a proper BC layer. That's why I wasn't sure why you think one option is good, and the other isn't, as both of them are equally good with a BC layer and upgrade docs in place.

iherak:
My comment was directed at 3.5, BC break is fine 4.0 :)

Misunderstanding was about the BC layer, with that included, I agree with you.

hknezevic:
I would rather have a new set of get* methods added now, and the old load* method set deprecated, then removed completely in 4.0.

I don't like the idea of having method arguments that can accept different data types. We are trying to move away from this approach since the times of eZ legacy, and I don't see how we would benefit from it now. Doesn't it make more sense to deprecate the method and not the parameter type?

emodric:
> I don't like the idea of having method arguments that can accept different data types. We are trying to move away from this approach since the times of eZ legacy, and I don't see how we would benefit from it now.

3.x methods already have different types in `load*` methods: `int|string $contentId`. Adding `Content` to that list is a BC layer, fairly standard when you look at Symfony BC policies, for example. What matters is that in 4.0, we're leaving only `Content` as a possible type, and we can add a proper typehint to method signature.

> Doesn't it make more sense to deprecate the method and not the parameter type?

Both of those make sense in equal way. I think it boils down to what method name you like more and would like to keep.

pspanja:
Changed as suggested, in order to keep the existing method names, see https://github.com/netgen/ezplatform-site-api/pull/173/commits/bc25873a54bbf320221c5ac8c0900d4bdde718bc

pspanja:
Thanks for the review.

PR #174, opened 2020-09-09
NGSTACK-423: preview shows published version instead of a draft

pspanja:
The problem was in the to eZ Platform view fallback template, which used Repository Content/Location. Since Repository Location lazy-loads it's Content and it has not information about the preview version, published version was displayed and rendered.

The fix simply passes Site Content/Location instead, which is already handled by the Twig extension. This should be sufficient as the function probably be used for full view only in automatic fallback.

Additionally the preview controller checked if the Location is draft, this seems unnecessary and is here removed for always mapping the Location from the existing data, including necessary `VersionInfo`.

One additional problem was found out while testing the preview, where when automatic fallback with subrequest is configured, links are not correctly generated for the selected siteaccess. This turned out to be a limitation of reverse siteaccess matching, where matcher in reversely matched siteaccess is not fully configured with all necessary data. This later propagates to fragment renderer through serialization and causes problems for a subrequest. Fixing this would have to involve fixing siteaccess matchers, which is out of our domain. A warning about this is added to the configuration docs.

FWIW, generating links works correctly without a subrequest, and this is already planned to be the default in v4.


emodric (approved):

pspanja:
Thanks @emodric

PR #176, opened 2020-09-14
NGSTACK-436: support Visible criterion from netgen/ezplatform-search-extra

pspanja:
This implements support for `Visible` criterion from `netgen/ezplatform-search-extra` (added in https://github.com/netgen/ezplatform-search-extra/pull/41). The criterion is used for Location children and siblings, Content relations and in Query Types.

With `Visible` criterion, Content is defined as visible if it's not hidden, and Location is defined as visible if it's not hidden by itself AND it's ancestor Locations AND it's Content is not hidden.

A bundle (semantic) configuration is implemented to control the default behaviour (showing the default):

```yaml
ezpublish:
    system:
        frontend_group:
            ng_site_api:
                show_hidden_items: false
```

This is added in order to configure default behaviour per siteaccess, so for example you can configure that hidden items are shown in admin UI context. The setting will affect fetching Location children and siblings, Content relations and in Query Types.

While no parameter is provided to override configured behaviour for Location children and siblings and Content relations, Query Types can override it by explicitly defining `visibile` condition, with values `true` (only visible), `false` (only hidden) and `null` (both visible and hidden).


docs/reference/query_types/parameters/common/content/visible.rst.inc
```diff
@@ -0,0 +1,29 @@
+``visible``
+~~~~~~~~~~~
+
+Defines whether returned Content objects are visible or not.
+Use ``true`` to get visible Content objects, ``false`` to get hidden Content objects and ``null`` to get both (which is also the default behaviour).
```
emodric: Should default be to get only visible objects?

pspanja: It actually is, I need to update the docs and also mention how configuration affects the default.

emodric: > Defines whether returned Content objects are visible or not.

Maybe we can reformulate this? This way, it kinda implies that somehow we will set the loaded objects to be visible or not.

Maybe something like: Filters and returns Content objects based on their visibility. Use `true` for...

pspanja: Updated in https://github.com/netgen/ezplatform-site-api/pull/176/commits/46ba67744d7a423428dc1cf344263d8c0d454fde

It irks me a bit that we have a kind of dichotomy with mapping the configuration between`show_hidden_items` and `visible`. The alternative would be renaming `show_hidden_items`  to `show_only_visible_items` and reversing the logic, but not sure if worth it.

emodric: So why not name the option here e.g. `include_hidden` and make it the default `false`? In 99% of the cases, if you wish to change the option, it is to change it so you load hidden items together with visible ones, in which case, `include_hidden: true` would make sense as a config.

We don't support loading only hidden objects with semantic config, why should we support it here?

pspanja: Because Query Types are only a configuration wrapper around PHP search API, so I would keep it as close to it as possible. What you suggest kind of makes sense, but then we lose the link between the configuration option and criterion.

emodric: Naming is hard, yeah 😅

pspanja: I think this is actually not a big deal. As one is a high level default configuration, and the other is explicit configuration of a query, it should not be confusing for developers.

emodric (approved):

pspanja:
Thanks @emodric

PR #180, opened 2020-09-17
Add missing properties to ContentInfo and Location

pspanja:
This adds new properties from Repository counterparts, `ContentInfo::$isHidden` and `Location::$explicitlyHidden`.

Since with addition of Content visibility the meaning of `Location::$hidden` and `Location::$invisible` changed, to properly check if the `$location` is visible one needs to check that `(!$location->hidden && !$location->invisible)`. See https://github.com/ezsystems/ezplatform-kernel/blob/2b64f12724c29bfad54438bc6d8fb7ca75d69094/eZ/Publish/API/Repository/Values/Content/Location.php#L116-L136 

For that reason I also added `$isVisible` property to ContentInfo and Location. The meaning is then as follows:

- `$location->visible = !$location->hidden && !$location->invisible`
- `$contentInfo->visible = !$contentInfo->hidden`

This should also make the logic and code more understandable, since it avoids double negation when checking if something is visible. For example:

```twig
{% if location.visible %}
    {{ location.content.name }}
{%endif%}
```

Instead of:

```twig
{% if not location.invisible %}
    {{ location.content.name }}
{%endif%}
```


emodric:
Just to make sure (and this could be added to Site API docs too), it is now enough to check `$contentInfo->isVisible` or `$location->isVisible`?

emodric (approved):

pspanja:
Since it doesn't exist on Repository objects, I also added `$isVisible` to the Site API Content object. Docs are updated as suggested.

pspanja:
Thanks @emodric

PR #184, opened 2020-09-28
NGSTACK-442: implement internal Content view route config

pspanja:
There a several problems with having Internal eZ Platform Content view route (`_ez_content_view`) on a frontend siteaccess:

- it generates the selected view using default templates from eZ Platform kernel
- it might present a security risk
  - it allows rendering of Content that might not be intended to be directly available (eg. various configuration objects)
  - it allows choosing a view and a Location
- it does not redirect to the URL alias route

This implementes rendering of the route through Site API and a configuration to control whether the route will be active on a frontend siteaccess and how it will behave (showing defaults):

```yaml
ezpublish:
    system:
        frontend_group:
            ng_site_api:
                enable_internal_view_route: true
                redirect_internal_view_route_to_url_alias: true
```

The route will be enabled by default and will permanently redirect to the URL alias. If disabled, or if the URL alias can't be generated, 404 will be thrown. If enabled without redirect, it will render the Content through the Site API, instead of using default eZ templates.

### DIFF
--- a/bundle/EventListener/InternalContentViewRouteListener.php
+++ b/bundle/EventListener/InternalContentViewRouteListener.php
@@ -83,7 +83,9 @@ class InternalContentViewRouteListener implements EventSubscriberInterface
         }
 
         if (!$this->configResolver->getParameter('ng_site_api.enable_internal_view_route')) {
-            throw new NotFoundHttpException();
+            throw new NotFoundHttpException(
+                'Internal Content view route has been disabled, check your Site API configuration for: "ng_site_api.enable_internal_view_route"'
+            );
         }
 
         $event->setResponse($this->getResponse($request));
@@ -155,7 +157,7 @@ class InternalContentViewRouteListener implements EventSubscriberInterface
         try {
             return $this->router->generate(UrlAliasRouter::URL_ALIAS_ROUTE_NAME, $parameters);
         } catch (Exception $e) {
-            throw new NotFoundHttpException();
+            throw new NotFoundHttpException('URL alias could not be generated');
         }
     }
 }
--- a/bundle/EventListener/InternalContentViewRouteListener.php
+++ b/bundle/EventListener/InternalContentViewRouteListener.php
@@ -1,0 +1,153 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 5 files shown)


bundle/EventListener/InternalContentViewRouteListener.php
```diff
@@ -0,0 +1,156 @@
...
+        ];
+
+        $locationId = $request->attributes->get('locationId');
+
+        if ($locationId !== null) {
+            $attributes['locationId'] = (int) $locationId;
+        }
+
+        return $this->fragmentHandler->render(
+            new ControllerReference('ng_content:viewAction', $attributes)
```
emodric: This form is deprecated in Symfony 5.1 I think, so it should be `ng_content::viewAction`

bundle/EventListener/InternalContentViewRouteListener.php
```diff
@@ -0,0 +1,156 @@
...
+        $event->setResponse($this->getResponse($request));
+    }
+
+    private function getResponse(Request $request): Response
+    {
+        if ($this->configResolver->getParameter('ng_site_api.redirect_internal_view_route_to_url_alias')) {
+            return new RedirectResponse($this->generateUrlAlias($request));
+        }
+
+        return new Response($this->renderView($request));
```
emodric, line 100: Instead of manually rendering the view, can't you just do nothing and let it fallback to eZ, or did you wish to render the view with Site API to prevent using fallback templates from eZ?

bundle/EventListener/InternalContentViewRouteListener.php
pspanja: Updated in https://github.com/netgen/ezplatform-site-api/pull/184/commits/a5e6dc9d9607a2c99efd3c3057d2fedbbad1e926

bundle/EventListener/InternalContentViewRouteListener.php
pspanja: Yes, otherwise it would fall back on the default eZ templates, which is not useful.

pspanja:
Updated as discussed in private; now the internal route is by default enabled, and will permanently redirect to the URL alias.

emodric (approved):

pspanja:
Thanks @emodric

PR #187, opened 2020-11-06
NGSTACK-461 Query type improvements

RandyCupic:
Sometimes values for query type parameters are being fetched dynamically and, if the dynamic value is NULL, currently it's not possible to avoid adding that criterion. To solve this, this PR enables various parameters to have NULL as value, which will prevent adding that criterion to the Query.

Documentation and tests have been updated accordingly.

Also, this PR introduces new `split` function to be used in query parameter values (needed for eg. splitting comma-separated list of supported content type identifiers into array for content type parameter).

### DIFF
--- a/docs/reference/query_types/parameters/common/query/limit.rst.inc
+++ b/docs/reference/query_types/parameters/common/query/limit.rst.inc
@@ -2,7 +2,7 @@
 ~~~~~~~~~
 
 Defines the maximum number of items to return.
-If ``null`` is used as a value, the limit won't be set.
+If ``null`` is used as a value, the limit will be set to the default value.
 
 .. note:: This parameter will not be used if you execute the query from Twig using ``ng_query`` function.
           In that case ``Pargerfanta`` pager is used with semantic parameters ``page`` and ``max_per_page``.
--- a/docs/reference/query_types/parameters/common/query/limit.rst.inc
+++ b/docs/reference/query_types/parameters/common/query/limit.rst.inc
@@ -2,12 +2,13 @@
 ~~~~~~~~~
 
 Defines the maximum number of items to return.
+If ``null`` is used as a value, the limit won't be set.
 
 .. note:: This parameter will not be used if you execute the query from Twig using ``ng_query`` function.
           In that case ``Pargerfanta`` pager is used with semantic parameters ``page`` and ``max_per_page``.
           To execute the query directly use ``ng_raw_query`` Twig function instead.
 
-- **value type**: ``integer``
+- **value type**: ``integer``, ``null``
 - **value format**: ``single``
 - **required**: ``false``
 - **default**: ``25``
--- a/docs/reference/query_types/parameters/common/query/offset.rst.inc
+++ b/docs/reference/query_types/parameters/common/query/offset.rst.inc
@@ -2,7 +2,7 @@
 ~~~~~~~~~~
 
 Defines the offset for search hits, used for paging the results.
-If ``null`` is used as a value, the offset won't be set.
+If ``null`` is used as a value, the offset will be set to the default value.
 
 .. note:: This parameter will not be used if you execute the query from Twig using ``ng_query`` function.
           In that case ``Pargerfanta`` pager is used with semantic parameters ``page`` and ``max_per_page``.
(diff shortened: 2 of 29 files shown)


docs/reference/query_types/parameters/common/query/sort/body.rst.inc
```diff
@@ -18,6 +18,15 @@ Following sort clauses are available through string definition:
...
     :local:
 
+Parent sort clausses
+~~~~~~~~~~~~~~~~~~~~
+
+String ``parent`` enables sorting by settings on parent location in administration.
+
+.. code-block:: yaml
+
+sort: parent
```
pspanja: Missing indent, it's not formatting properly.

docs/reference/query_types/parameters/common/query/sort/body.rst.inc
```diff
@@ -18,6 +18,15 @@ Following sort clauses are available through string definition:
     :depth: 1
     :local:
 
+Parent sort clausses
```
pspanja: Parent sort clause*

lib/Core/Site/QueryType/Base.php
```diff
@@ -394,18 +401,36 @@ private function resolveFilterCriteria(array $parameters): ?Criterion
...
     private function getSortClauses(array $parameters): array
     {
+        $sortDefinitions = $parameters['sort'];
+        $sortClauses = [];
+
+        if (empty($sortDefinitions)) {
+            return $sortClauses;
+        }
+
+        if (\strtolower(\trim($sortDefinitions)) === 'parent' && array_key_exists('location', $parameters)) {
```
pspanja: Two issues with handling parent sort clause like this:

- `sort: parent` is in fact an alias for `sort: [parent]`, this makes combining with other sort clauses impossible
- will not give an error on incorrect usage (in Content query type)

I would propose to solve these by implementing it in the `SortClauseParser` service. If the Location can't be found, invalid argument exception should be thrown, and the existing mechanism will take care of throwing the exception in `self::parseCustomSortString()`.

Alternatively, it's possible to configure sort with `SortClause` instances, so we can also handle it through an expression:

```yaml
sort: '@=location.parent.innerLocation.getSortClauses()'
```

RandyCupic: @pspanja Sorry, I didn't know that it's possible to provide directly sort clauses. I reverted all the changes and I will use it with direct sort clauses.

docs/reference/query_types/parameters/common/query/sort/body.rst.inc
RandyCupic: @pspanja This change has been reverted.

docs/reference/query_types/parameters/common/query/sort/body.rst.inc
RandyCupic: @pspanja This change has been reverted.

docs/reference/query_types/parameters/common/query/offset.rst.inc
```diff
@@ -2,12 +2,13 @@
 ~~~~~~~~~~
 
 Defines the offset for search hits, used for paging the results.
+If ``null`` is used as a value, the offset won't be set.
```
pspanja: The query object still has `0` for the default offset, so while `null` won't be set on it, the resulting offset will still be `0`. That's not obvious, but it's what we should document, since that's what the user will end up with.

Not sure, but maybe normalizers can be used to convert `null` to `0`.

docs/reference/query_types/parameters/common/query/limit.rst.inc
```diff
@@ -2,12 +2,13 @@
 ~~~~~~~~~
 
 Defines the maximum number of items to return.
+If ``null`` is used as a value, the limit won't be set.
```
pspanja: Similar as commented for offset, with `null` limit will default to `25`.

docs/reference/query_types/parameters/common/query/offset.rst.inc
RandyCupic: Fixed in https://github.com/netgen/ezplatform-site-api/pull/187/commits/31e62e94e070fe6f8e0616ae050d6551fc4e2eb4

docs/reference/query_types/parameters/common/query/limit.rst.inc
RandyCupic: Fixed in https://github.com/netgen/ezplatform-site-api/pull/187/commits/31e62e94e070fe6f8e0616ae050d6551fc4e2eb4

pspanja:
Thanks @RandyCupic!

PR #190, opened 2020-11-28
NGSTACK-480: implement access to relation Locations

pspanja:
This implements access to relation Locations.

Note that this does not implement real Location relations, related Locations are are **visible** **main** Locations of the related Content, and relations are internally still implemented on Content level.

Two new methods are added to the `RelationService`:
- `loadFieldRelationLocation()`
- `loadFieldRelationLocations()`

Three new methods are added to the `Content` object:
- `getFieldRelationLocation()`
- `getFieldRelationLocations()`
- `filterFieldRelationLocations()`

Four new query types are implemented:
- `SiteAPI:Location/Relations/AllTagFields`
- `SiteAPI:Location/Relations/ForwardFields`
- `SiteAPI:Location/Relations/ReverseFields`
- `SiteAPI:Location/Relations/TagFields`


hknezevic (approved):
@pspanja looks good!

emodric:
Agreed, looks good :+1:

lib/API/RelationService.php
```diff
@@ -50,4 +51,42 @@ public function loadFieldRelations(
...
+     *
+     * Note: only visible main Location of the related Content will be used.
+     *
+     * @param \Netgen\EzPlatformSiteApi\API\Values\Content $content
+     * @param string $fieldDefinitionIdentifier
+     * @param array $contentTypeIdentifiers
+     *
+     * @return null|\Netgen\EzPlatformSiteApi\API\Values\Location
+     */
+    public function loadLocationFieldRelation(
```
iherak: Naming nitpick:
I'd prefer this to be named `loadFieldRelationLocation` as we are loading location(s) from the field relation.
Current naming implies that we are loading field relation from the location, which is not true.

Same for all other corresponding methods.

I am not sure if I am making sense here, so feel free to ask for more explanations :)

hknezevic: @iherak Actually, now that you pointed it out, I agree it does make more sense with the naming pattern you proposed. I didn't give it much thought at the first glance.
So the desired format change would be:
loadLocationFieldRelation(s) -> loadFieldRelationLocation(s)
filterLocationFieldRelation(s) -> filterFieldRelationLocation(s)
cc @pspanja

pspanja: Agreed, that's much better 👍

pspanja: Updated in https://github.com/netgen/ezplatform-site-api/pull/190/commits/8a63631cddb0d5b502f1296da922012cd9b45c86

iherak (approved):
Looks great now!

pspanja:
Thanks for the review!

PR #194, opened 2020-12-14
[3.7] Migrate Travis to GitHub Actions

pspanja:


emodric:
Hooray :tada:

pspanja:
This was easier than expected, thanks for the examples on the layouts repo 👍

emodric:
Yep, my only gripe with it is that, short of writing your own actions, you can't share jobs between steps, or steps between workflows, resulting in different workflows/steps repeating not-so-small amounts of commands.

PR #201, opened 2021-07-29
Fix wrong condition in code examples for location priority

bcafuk:
The code snippets on the page for locations' _priority_ condition contain mistakes: The examples don't actually use the `priority` condition. They currently use the `depth` condition instead.

This pull requests replaces uses of the `depth` condition with the `priority` condition being documented.


pspanja (approved):

pspanja:
Thanks @bcafuk

Issue #204, opened 2021-09-16
Make bundle compatible with sensio/framework-extra-bundle ^6.0


rembertin:
Currently, the bundle requires `sensio/framework-extra-bundle` at version `^5.5`.
The problem is that since its version `3.3.6`, the package [ibexa/oss](https://packagist.org/packages/ibexa/oss) requires `sensio/framework-extra-bundle` at version `^6.1.0`.
So currently it's not possible to install `netgen/ezplatform-site-api` with the last version of Ibexa DXP.

Could it be possible to require `sensio/framework-extra-bundle` either at version `^5.5` and `^6.0`? (in composer.json: `"sensio/framework-extra-bundle": "^5.5|^6.0"`).

Thanks in advance.

pspanja:
Hi @rembertin, thanks for the report. I’m on vacation ATM, so will look into it in the next couple of weeks.

rembertin:
Hi, thanks for your answer. I created a PR for this if it can make your life easier. => #205
Have a nice vacation!

pspanja:
Fixed in #205

PR #205, opened 2021-09-17
Require sensio/framework-extra-bundle at version ^5.5 or ^6.0

rembertin:
Q | A
-- | --
Branch? | master
Bug fix? | no
New feature? | yes
BC breaks? | no
Deprecations? | no
Tests pass? | yes
Fixed tickets? | #204

The purpose is to make this bundle compatible with the last version of Ibexa DXP, which requires `sensio/framework-extra-bundle` at version `^6.1`.

It seems that the only usage of `sensio/framework-extra-bundle` is for param converters, so there should not be any BC breaks.

This is my first PR ever, so don't hesitate to correct me if I'm wrong on the process. :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "ezsystems/ezplatform-http-cache": "^2.2",
         "ezsystems/ezplatform-richtext": "^2.2",
         "netgen/ezplatform-search-extra": "^2.5",
-        "sensio/framework-extra-bundle": "^5.5"
+        "sensio/framework-extra-bundle": "^5.5|^6.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^9.0",


pspanja:
Hi @rembertin, I tried testing this with OSS and it works OK, so wondering is the Ibexa DXP you use a closed source version?

rembertin:
No, I use the public `ibexa/oss` package (https://github.com/ibexa/oss). In fact, since `ibexa/oss` v3.3.6, the package requires `sensio/framework-extra-bundle:^6.1.0`, that's why I created this PR.

pspanja:
Go it now, thanks. This looks good to go :)

pspanja:
Thanks @rembertin 👍 

Released as [4.3.2](https://github.com/netgen/ezplatform-site-api/releases/tag/4.3.2).

rembertin:
Thank you very much!

PR #206, opened 2021-11-02
Improve templating.rst

mlukac:
Checking for field relation if it's not empty seems better, "is defined" breaks the page by returning null if there is no related object.

### DIFF
--- a/docs/reference/templating.rst
+++ b/docs/reference/templating.rst
@@ -413,7 +413,7 @@ Content Field relations
 
     {% set related_content = content.fieldRelation('related_article') %}
 
-    {% if related_content is defined %}
+    {% if related_content is not empty %}
         <a href="{{ ez_path(related_content) }}">{{ related_content.name }}</a>
     {% else %}
         <p>There are two possibilities:</p>


docs/reference/templating.rst
```diff
@@ -413,22 +413,16 @@ Content Field relations
...
-        </ol>
-        <p>In any case, you can't render the related Content!</p>
     {% endif %}
 
   .. note::
 
     If relation field contains multiple relations, the first one will be returned. If it doesn't
     contain relations or you don't have the access to read the related Content, the method will
-    return ``null``. Make sure to check if that's the case.
+    return ``
```
iherak: Is this accidental change or?

docs/reference/templating.rst
```diff
@@ -413,22 +413,16 @@ Content Field relations
 
     {% set related_content = content.fieldRelation('related_article') %}
 
-    {% if related_content is defined %}
+    {% if related_content is not empty %}
         <a href="{{ ez_path(related_content) }}">{{ related_content.name }}</a>
-    {% else %}
```
iherak, line 418: Why should we remove this? This provides context to the person reading the docs?

docs/reference/templating.rst
mlukac: Yes, accident.

docs/reference/templating.rst
mlukac: Yes, but text needs changes. Do you have suggestions how to edit these tips? Is the first posibility is redundat now. Don't know if second posibility is valid with new condition?

iherak: Fixed through amend.

docs/reference/templating.rst
iherak: Fixed through amend.

iherak (approved):

emodric:
Thanks @mlukac !

